PR #348, opened 2023-02-27
[extension:chrome,edge] update default storageType to "local"

sagarsoni7:


lazyvab (approved):

lazyvab:
@sagarsoni7 In future, we should also refactor MV2 code to hard-code storage type to local instead of reading from config.

PR #357, opened 2023-03-01
[DNM] Promote Changes to Prod

lazyvab:


lazyvab:
This got accidentally merged while releasing extension from prod.
Here is the new PR: https://github.com/requestly/requestly/pull/363

Issue #433, opened 2023-03-23
Modify API Response rule fails to intercept requests in few scenarios


lazyvab:
Environment: Chrome extension

When an API request is triggered early in the page lifecycle, the Requestly interceptors fail to attach by then and the response could not be modified.

This issue has recently crept in after https://github.com/requestly/requestly/issues/338 was fixed.

PR #434, opened 2023-03-23
#433 Modify API Response rule fails to intercept requests in few scenarios

lazyvab:
Closes issue: [](https://github.com/requestly/requestly/issues/433)

## 📜 Summary of changes:

- The client content script will not query the response rule from ID directly, but the data will be passed down from background script. It saves an async storage request in client.
- The client content script will not check for extension status during initialization. This check used to delay the setup of different rule handlers. It will now be checked in individual flows while execution.

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/browser-extension/mv2/src/client/js/client.js
+++ b/browser-extension/mv2/src/client/js/client.js
@@ -3,25 +3,14 @@
     return;
   }
 
-  chrome.runtime.sendMessage(
-    {
-      action: RQ.EXTENSION_MESSAGES.CHECK_IF_EXTENSION_ENABLED,
-    },
-    (isExtensionEnabled) => {
-      if (!isExtensionEnabled) {
-        return;
-      }
+  RQ.ConsoleLogger.setup();
+  RQ.RuleExecutionHandler.setup();
+  RQ.ScriptRuleHandler.setup();
+  RQ.SessionRecorder.setup();
+  RQ.UserAgentRuleHandler.setup();
+  RQ.RequestResponseRuleHandler.setup();
 
-      RQ.ConsoleLogger.setup();
-      RQ.RuleExecutionHandler.setup();
-      RQ.ScriptRuleHandler.setup();
-      RQ.SessionRecorder.setup();
-      RQ.UserAgentRuleHandler.setup();
-      RQ.RequestResponseRuleHandler.setup();
-
-      chrome.runtime.sendMessage({
-        action: RQ.CLIENT_MESSAGES.NOTIFY_CONTENT_SCRIPT_LOADED,
-      });
-    }
-  );
+  chrome.runtime.sendMessage({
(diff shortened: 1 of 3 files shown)


browser-extension/mv2/src/client/js/client.js
```diff
@@ -3,25 +3,14 @@
...
-    (isExtensionEnabled) => {
-      if (!isExtensionEnabled) {
-        return;
-      }
+  RQ.ConsoleLogger.setup();
+  RQ.RuleExecutionHandler.setup();
+  RQ.ScriptRuleHandler.setup();
+  RQ.SessionRecorder.setup();
+  RQ.UserAgentRuleHandler.setup();
+  RQ.RequestResponseRuleHandler.setup();
```
sachinjain024, line 11: @lazyvab I think it was fair to have all the handlers being initialized on the basis of extension is enabled or not. Do you think this is also causing enough delay for creating race conditions for AJAX interception?

browser-extension/mv2/src/client/js/requestResponseRuleHandler.js
```diff
@@ -3,6 +3,19 @@ RQ.RequestResponseRuleHandler.isInitialized = false;
 RQ.RequestResponseRuleHandler.cachedResponseRuleIds = new Set();
 
 RQ.RequestResponseRuleHandler.setup = () => {
+  // TODO: store the extension status in background and watch for changes here
```
sachinjain024, line 6: Extension status is already cached/available in memory to background but I don't think we should watch for changes here. Instead content script should ask background whenever it needs to know the extension status on demand but I guess you are doing this to avoid an async operation here

browser-extension/mv2/src/client/js/requestResponseRuleHandler.js
```diff
@@ -81,16 +94,14 @@ RQ.RequestResponseRuleHandler.cacheRequestRules = () => {
...
-    RQ.RequestResponseRuleHandler.cachedResponseRuleIds.add(ruleId);
-  });
+RQ.RequestResponseRuleHandler.cacheResponseRule = (url, responseRule) => {
+  RQ.ClientUtils.executeJS(
+    `window.${
+      RQ.PUBLIC_NAMESPACE
+    }.responseRulesByUrl['${url}'] = ${JSON.stringify(responseRule)};`,
+    true
+  );
+  RQ.RequestResponseRuleHandler.cachedResponseRuleIds.add(responseRule.id);
```
sachinjain024, line 104: Why do we need to maintain cachedResponseRuleIds?

sachinjain024 (commented):
Added a couple of comments, mainly asking for clarification

sachinjain024 (approved):

PR #454, opened 2023-03-29
chore: template code for batched event writer

nsrCodes:


browser-extension/mv2/src/background/background.js
```diff
@@ -1,3 +1,75 @@
+// import { startEventBatchWriter } from "./EventWriter";
+
+/* can be moved to a different file --- EventWriter.js */
+
+let eventsBatch = [];
```
nafees87n, line 5: Lets name this `events` or `analyticEvents`. We shouldn't use the word `batch` unless batches are created. Its confusing

browser-extension/mv2/src/background/background.js
```diff
@@ -1,3 +1,75 @@
...
+let batchWriterInterval = null;
+
+const LOCAL_ANALYTICS_KEY = "analytics_event";
+
+function addEventToBatch(event) {
```
nafees87n, line 10:
```suggestion
function addEvent(event) {
```
or anything relevant

browser-extension/mv2/src/background/background.js
```diff
@@ -1,3 +1,75 @@
...
+      await writeBatchToLocalStorage();
+    }, intervalTime);
+  }
+
+  return batchWriterInterval;
+}
+
+async function getEvents() {
+  // export
+  const events = RQ.StorageService.getRecord(LOCAL_ANALYTICS_KEY)
```
nafees87n, line 57:
```suggestion
  const events = 
await RQ.StorageService.getRecord(LOCAL_ANALYTICS_KEY)
```
or should use try-catch

browser-extension/mv2/src/background/background.js
```diff
@@ -1,3 +1,75 @@
...
+  // todo: generate uuid for each event
+  eventsBatch.push(event);
+}
+
+async function writeBatchToLocalStorage() {
+  const batchedEvents = [...getBatchedEvents()];
+
+  if (batchedEvents.length) {
+    // write batched events to local storage
+    await RQ.StorageService.getRecord(LOCAL_ANALYTICS_KEY).then(
```
nafees87n, line 21: isn't `await` unnecessary here?

nafees87n, line 21: `getEvents()` can be used here

nafees87n (changes requested):
Lets use this convention:
when the events haven't been written to the written to the storage, lets call it `events` and the events present in storage as `batchedEvents`.

Then we'll have 2 arrays now:
1. Events in local array: `events[]`
2. Batched events in storage: `batchedEvents[]`- in future when amount of events will increase, we'll have a map instead -
```
batchedEvents={
  [batchId]:events[]
}
```

browser-extension/mv2/src/background/background.js
nsrCodes: can't see any difference either way, we just need a way to return null or empty array when events are not stored.
Both work the same IMO so not changing for now. 
as for awaiting, this returns the promise to the main action of the function so for now, again doesn't make a difference

browser-extension/mv2/src/background/background.js
```diff
@@ -2,27 +2,27 @@
...
-let eventsBatch = [];
-let batchWriterInterval = null;
+let eventsToWrite = [];
+let eventWriterInterval = null;
 
 const LOCAL_ANALYTICS_KEY = "analytics_event";
 
 // to be exported
-function addEventToBatch(event) {
+function queueEventForWrite(event) {
```
nafees87n, line 11:
```suggestion
function queueEventToWrite(event) {
```

browser-extension/mv2/src/background/background.js
```diff
@@ -2,27 +2,27 @@
...
 
-async function writeBatchToLocalStorage() {
-  const batchedEvents = [...getBatchedEvents()];
+async function writeEventsToLocalStorage() {
+  const localEventsToWrite = [...getEventsToWrite()];
 
-  if (batchedEvents.length) {
+  if (localEventsToWrite.length) {
     // write batched events to local storage
     return getEvents().then(async (storageEvents) => {
```
nafees87n, line 21:
```suggestion
    return getEvents().then((storageEvents) => {
```

nafees87n (approved):
Merge after these

PR #455, opened 2023-03-29
extension events data flow pipeline

nafees87n:


browser-extension/mv2/src/background/background.js
```diff
@@ -974,6 +974,61 @@ BG.Methods.sendMessage = function (messageObject, callback) {
...
+  }
+
+  if (!BG.isAppOnline) {
+    return;
+  }
+
+  const extensionEventsMessage = {
+    action: RQ.EXTENSION_MESSAGES.EXTENSION_EVENTS,
+    events: {}, // it should fetched from a separate function only after all the checks and shouldn't be passed as params
+  };
```
nsrCodes, line 1017: We will not get an action from anywhere, will be triggered by a setInterval running in the background script itself

reference code around this commit -- https://github.com/requestly/requestly/pull/454/commits/651785841f82085b8a39d9be0392a583d63fb927

browser-extension/mv2/src/background/background.js
```diff
@@ -974,6 +974,61 @@ BG.Methods.sendMessage = function (messageObject, callback) {
...
+      if (res.received) {
+        console.log("!!!debug", "ack in bg!! clear the events!!");
+      }
+    })
+    .catch(() => {
+      if (retryCounter < 3) {
+        BG.Methods.sendExtensionEvents(++retryCounter);
+      }
+    });
+};
```
nsrCodes, line 1030: unnecessary function, wasn't mentioned in initial architecture discussions

browser-extension/mv2/src/background/background.js
```diff
@@ -974,6 +974,61 @@ BG.Methods.sendMessage = function (messageObject, callback) {
...
+
+  return Promise.race([
+    sendMessage(messageObject, tabId),
+    new Promise((_, reject) =>
+      setTimeout(() => reject(new Error("timeout")), timeout)
+    ),
+  ]);
+};
+
+BG.Methods.sendExtensionEvents = (tabId = null, retryCounter = 0) => {
```
nsrCodes, line 998: This function will not have any arguments for now

browser-extension/mv2/src/background/background.js
```diff
@@ -974,6 +974,61 @@ BG.Methods.sendMessage = function (messageObject, callback) {
...
+  let appTabId = tabId;
+  if (!appTabId) {
+    chrome.tabs.query({ url: RQ.configs.WEB_URL + "/*" }, (tabs) => {
+      if (tabs.length > 0) {
+        appTabId = tabs[0].id;
+      } else {
+        BG.isAppOnline = false;
```
nsrCodes, line 1005: return after this and remove the extra if statement below

browser-extension/mv2/src/background/background.js
```diff
@@ -974,6 +974,61 @@ BG.Methods.sendMessage = function (messageObject, callback) {
...
+    sendMessage(messageObject, tabId),
+    new Promise((_, reject) =>
+      setTimeout(() => reject(new Error("timeout")), timeout)
+    ),
+  ]);
+};
+
+BG.Methods.sendExtensionEvents = (tabId = null, retryCounter = 0) => {
+  let appTabId = tabId;
+  if (!appTabId) {
```
nsrCodes, line 1000: should be `if(BG.isAppOnline)`

browser-extension/mv2/src/background/background.js
```diff
@@ -974,6 +974,61 @@ BG.Methods.sendMessage = function (messageObject, callback) {
   );
 };
 
+BG.Methods.sendMessageToApp = (messageObject, tabId, timeout = 2000) => {
+  const sendMessage = (messageObject, tabId) => {
```
nsrCodes, line 978: This function will not get the message object from anywhere

As discussed, this is not a reusable function for now and will also be responsible for reading events from local storage

so you still have to create a separate function called `sendEventsToTab` that does all this and also the batching of events to handle the edge cases / acknowledgement procedure of this function

browser-extension/mv2/src/background/background.js
```diff
@@ -974,6 +974,61 @@ BG.Methods.sendMessage = function (messageObject, callback) {
...
+    return new Promise((resolve, reject) => {
+      chrome.tabs.sendMessage(tabId, messageObject, (response) => {
+        if (response) {
+          resolve(response);
+        } else {
+          reject(new Error("No response from app"));
```
nsrCodes, line 984: is this reject required?

nsrCodes (changes requested):

browser-extension/mv2/src/background/background.js
nafees87n: No, we anyway need a `sendMessageToApp` which will be a general, reusable function. Whatever event function we create, internally in the last step, it must call `sendMessageToApp` to send the event to App/UI.

browser-extension/mv2/src/background/background.js
nsrCodes: remove reject, will be covered by timeout

browser-extension/mv2/src/modules/EventActions.js
```diff
@@ -0,0 +1,48 @@
+const EventActions = {};
+
+EventActions.sendExtensionEvents = async () => {
+  if (!BG.isAppOnline) {
+    return;
+  }
+
+  let ackReceived = false;
```
nsrCodes, line 8: no need to save this as a variable, 

we can just break the loop based on the response from `sendMessageToApp`

nsrCodes (changes requested):

PR #476, opened 2023-04-04
fix: extension runtime message errors

nafees87n:
## 📜 Summary of changes:
- fixes the errors in extension runtime messages
- send message to sync applied rules in content script only when rules are cached

### DIFF
--- a/browser-extension/mv2/src/client/js/ruleExecutionHandler.js
+++ b/browser-extension/mv2/src/client/js/ruleExecutionHandler.js
@@ -18,7 +18,8 @@ RQ.RuleExecutionHandler.setup = () => {
           message.appliedRuleDetails,
           message.isConsoleLoggerEnabled
         );
-        break;
+        sendResponse(true);
+        return true;
     }
     return false;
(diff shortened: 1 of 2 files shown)


browser-extension/mv2/src/client/js/ruleExecutionHandler.js
```diff
@@ -18,7 +18,8 @@ RQ.RuleExecutionHandler.setup = () => {
           message.appliedRuleDetails,
           message.isConsoleLoggerEnabled
         );
-        break;
+        sendResponse(true);
```
lazyvab, line 21: Do you need to send `true` as argument?

lazyvab (approved):

browser-extension/mv2/src/client/js/ruleExecutionHandler.js
nafees87n: No, I dont need to send `true` but we are using a callback in the sender function so a response is expected. Hence, we were getting the error.

PR #480, opened 2023-04-05
RQ-97 Extension Analytics - Fixed Cyclic dependency issue

lazyvab:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.


app/src/config/PageScriptMessageHandler.js
```diff
@@ -86,7 +86,13 @@ const PageScriptMessageHandler = {
     ) {
       Logger.log("Received message:", event.data);
 
-      if (event.data.requestId) {
+      if (
+        event.data.requestId &&
```
lazyvab, line 90: @nafees87n If this receives a message with `requestId` defined, can't we consider it as a response case and invoke the callback? Or in some cases, we would expect `messageHandler` to handle it?

Also, the existence of action_requestId is already handled in `invokeCallback` function.

nafees87n (approved):

PR #538, opened 2023-04-14
feat: rule execution events

nsrCodes:


browser-extension/mv2/src/client/js/ruleExecutionHandler.js
```diff
@@ -6,30 +6,35 @@ RQ.RuleExecutionHandler.setup = () => {
...
+        sendResponse(isFirstExecution);
         break;
 
       case RQ.CLIENT_MESSAGES.GET_APPLIED_RULE_IDS:
         sendResponse(Array.from(RQ.RuleExecutionHandler.appliedRuleIds));
         break;
 
       case RQ.CLIENT_MESSAGES.SYNC_APPLIED_RULES:
-        RQ.RuleExecutionHandler.syncCachedAppliedRules(
+        const newRules = RQ.RuleExecutionHandler.syncCachedAppliedRules(
```
nafees87n, line 20: no need for this change. cachedRules for toplevel redirect are present in background itself.

browser-extension/mv2/src/client/js/ruleExecutionHandler.js
```diff
@@ -6,30 +6,35 @@ RQ.RuleExecutionHandler.setup = () => {
...
     return false;
   });
 };
 
-RQ.RuleExecutionHandler.syncCachedAppliedRules = (
-  appliedRuleDetails,
-  isConsoleLoggerEnabled
-) => {
+RQ.RuleExecutionHandler.syncCachedAppliedRules = (appliedRuleDetails, isConsoleLoggerEnabled) => {
+  const newRules = [];
```
nafees87n, line 32: No need for any change in this function

browser-extension/mv2/src/modules/EventActions.js
```diff
@@ -186,3 +179,22 @@ EventActions.stopPeriodicEventWriter = () => {
...
+  RULE_EXECUTED: "rule_executed",
+};
+
+EventActions.sendRuleExecutionEvent = (rule) => {
+  const eventParams = {
+    rule_type: rule.ruleType,
+    rule_id: rule.id,
+  };
+
+  eventParams["log_source"] = "extension";
```
nafees87n, line 194: `source` is not present in `eventParams`

nafees87n (changes requested):
Please see the suggested changes. The rules applied on top level document are cached in background. So it can be fetched from there itself.
Moreover, the changes have broken the logging of cached rules somehow.

browser-extension/mv2/src/modules/EventActions.js
lazyvab: @nafees87n @nsrCodes Instead of `source`, can we rather set `platform=extension`. Other values of platform - desktop, mobile, etc. 
On a single platform, there will always be one source. That is, `background` in `extension`, `proxy` in `desktop`, `sdk` in `mobile`. So, we can skip `source`. WDYT?

nsrCodes: @lazyvab as discussed earlier, IMO `source` sounded confusing as currently we use it for a very different context in web app events. I would like using `platform` instead. 

So just to be clear, `platform` would replace `log_source` and `source` would remain as is.

nafees87n (approved):

browser-extension/mv2/src/background/background.js
```diff
@@ -1442,29 +1241,28 @@ BG.Methods.saveExecutionCount = async function (rule) {
...
-    action: RQ.CLIENT_MESSAGES.UPDATE_APPLIED_RULE_ID,
-    ruleId: rule.id,
-  });
+  chrome.tabs.sendMessage(
+    tabId,
+    {
+      action: RQ.CLIENT_MESSAGES.UPDATE_APPLIED_RULE_ID,
+      ruleId: rule.id,
+    },
+    (isFirstExecution) => {
```
lazyvab, line 1256: Can we rather send full rule definition to client in the message and let the content script trigger the event similar to popup and devtool?
In that case, the action can be renamed to `NOTIFY_RULE_APPLIED`.

browser-extension/mv2/src/client/js/ruleExecutionHandler.js
```diff
@@ -2,11 +2,30 @@ RQ.RuleExecutionHandler = {
...
 };
 
+RQ.RuleExecutionHandler.sendRuleExecutionEvent = (rule) => {
+  const eventName = "rule_execution";
+  const eventParams = {
+    rule_type: rule.ruleType,
+    rule_id: rule.id,
+    platform: "extension",
+  };
+  eventParams["source"] = "content_script";
```
nafees87n, line 12: No need for source. `source` does not represent the source of the event here. `platform` is apt for `rule_execution`

browser-extension/mv2/src/client/js/ruleExecutionHandler.js
```diff
@@ -2,11 +2,30 @@ RQ.RuleExecutionHandler = {
   appliedRuleIds: new Set(),
 };
 
+RQ.RuleExecutionHandler.sendRuleExecutionEvent = (rule) => {
+  const eventName = "rule_execution";
```
nafees87n, line 6: Should we name it `rule_executed`?

nafees87n (changes requested):

browser-extension/mv2/src/client/js/ruleExecutionHandler.js
nsrCodes: I don't know, doesn't hurt to add it I guess.

browser-extension/mv2/src/client/js/ruleExecutionHandler.js
```diff
@@ -25,11 +44,11 @@ RQ.RuleExecutionHandler.setup = () => {
 RQ.RuleExecutionHandler.syncCachedAppliedRules = (appliedRuleDetails, isConsoleLoggerEnabled) => {
   appliedRuleDetails.forEach((appliedRule) => {
     RQ.RuleExecutionHandler.appliedRuleIds.add(appliedRule.rule.id);
-
     RQ.ConsoleLogger.handleMessage({
       requestDetails: appliedRule.requestDetails,
       rule: appliedRule.rule,
       isConsoleLoggerEnabled,
     });
+    RQ.RuleExecutionHandler.sendRuleExecutionEvent(appliedRule.rule);
```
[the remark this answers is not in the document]
nsrCodes: doesn't make sense to make a separate function for a simple functionality. But will add a check here too

nafees87n (approved):

Issue #622, opened 2023-05-17
Set request initiator origin  in access-control-allow-origin header to solve CORS issue


lazyvab:
To bypass CORS issue for a request in browser, `access-control-allow-origin` response header can be set as `*` using `Modify Headers` rule. But it does not work in some cases like when the request's credentials mode is 'include'.

Looking for a solution where the domain of the page can be populated as the header value, i.e.
```
When request URL contains "localhost",
Add response header "Access-Control-Allow-Origin" = {{page_domain}}
```

![image](https://github.com/requestly/requestly/assets/6367566/46391467-fe1d-4e75-90c5-8ee1efcd43c9)

lazyvab:
In case a request originates from an iframe, the value of `access-control-allow-origin` response header should be origin derived from the iframe URL, instead of main page URL.

So, going to expose a new predefined function `rq_request_initiator_origin()` which can be used to get desired value.

lazyvab:
Released in extension version `23.5.19`.

PR #628, opened 2023-05-18
#622: Set request initiator origin in access-control-allow-origin header to solve CORS issue

lazyvab:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.


browser-extension/mv2/tests/specs/modules/PreDefinedFunctions.spec.js
```diff
@@ -109,25 +81,52 @@ describe("PreDefined Functions: ", function () {
...
       expect(F.eval("rq_decrement(10, 3)")).toBe("7");
       expect(F.eval("rq_decrement(10 ,3)")).toBe("7");
       expect(F.eval("rq_decrement(10 , 3)")).toBe("7");
-      expect(F.eval("http://www.example.com/?page=rq_decrement(10, 3)")).toBe(
-        "http://www.example.com/?page=7"
-      );
+      expect(F.eval("http://www.example.com/?page=rq_decrement(10, 3)")).toBe("http://www.example.com/?page=7");
+    });
+  });
+
```
lazyvab, line 99: Below this line is the actual addition in this file. Rest are linting fixes pushed automatically with commit hook.

browser-extension/mv2/src/modules/PreDefinedFunctions.js
```diff
@@ -80,3 +76,25 @@ RQ.registerPredefinedFunction("rq_decrement", {
     return parseInt(num) - parseInt(step);
   },
 });
+
+RQ.registerPredefinedFunction("rq_request_origin", {
```
sachinjain024, line 80: @lazyvab request_origin can be confused with the origin of the request URL. What we want here is to define a predefined function that returns tab_origin or page_origin So should we rename this predefined function to `rq_page_origin`

sachinjain024:
Can we also update the CORS example template as part of this pull request only? Or, Do we want to take that up later?

browser-extension/mv2/src/modules/PreDefinedFunctions.js
lazyvab: @sachinjain024 The request origin is always defined as the origin which initiated the request. The `origin` header included in a request by the browser also means the same thing.

Also, in case of CORS issue, browser console log looks like:
`Access to fetch at 'https://mysite.com/test-api' from origin 'https://example.com' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header has a value 'https://mysite.com' that is not equal to the supplied origin.`

If you think it can still be confusing, can we rename to `rq_request_initiator`?

The reason why I don't want to call it `page_origin` is because if the request is initiated from an iframe, the iframe's origin might be different from page_origin.

lazyvab: As discussed, we are renaming it to `rq_origin()`.

lazyvab:
> Can we also update the CORS example template as part of this pull request only? Or, Do we want to take that up later?

Updated in latest commit in this PR.

PR #641, opened 2023-05-24
RQ-365: Remove system wide proxy CTA when it is not activateable

nsrCodes:


app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/NetworkInspector/NetworkTable/NoTrafficCTA.jsx
```diff
@@ -180,8 +180,7 @@ const NoTrafficCTA = ({ isStaticPreview }) => {
     navigate,
     numberOfConnectedApps,
     openConnectedAppsModal,
-    systemWideSource.isActive,
-    systemWideSource.isAvailable,
+    systemWideSource,
```
rohanmathur91, line 183: Put `systemWideSource.isActive` and `systemWideSource.isAvailable,` in dependency array instead of whole object, if sure that we will get stable `systemWideSource` object for each rerender then its okay.

rohanmathur91 (approved):

PR #684, opened 2023-06-05
Add Arc browser to list connected apps

nsrCodes:
Will be compatible once the supporting PR is released for desktop app: https://github.com/requestly/requestly-desktop-app/pull/37


app/src/config/constants/compatibility.js
```diff
@@ -92,4 +92,9 @@ export const FEATURE_COMPATIBLE_VERSION = {
     [GLOBAL_CONSTANTS.APP_MODES.EXTENSION]: "23.5.9",
     [GLOBAL_CONSTANTS.APP_MODES.REMOTE]: null,
   },
+  [FEATURES.ARC_BROWSER]: {
+    [GLOBAL_CONSTANTS.APP_MODES.DESKTOP]: null,
+    [GLOBAL_CONSTANTS.APP_MODES.EXTENSION]: null,
+    [GLOBAL_CONSTANTS.APP_MODES.REMOTE]: null,
+  },
```
wrongsahil, line 99: Can't we handle this without compatibility 🤔? If arc supports doesn't exist in the background process, then it just sends not supported response.

wrongsahil (changes requested):

app/src/config/constants/compatibility.js
nsrCodes: yes you are right. In fact, currently, no extra logic is needed for checking app support. The provided app will just **not** be responded as `activateable` by the background process.



The reason for adding this is to keep the new UI compatible with older versions. 
I just wasn't sure if the behaviour of older versions is error-free in the case when the app supplied by UI is not being managed in the background.

nsrCodes: removed compatibility check, they are just not needed

wrongsahil (approved):

PR #707, opened 2023-06-10
chore: show connect app modal on start if user has never connected any traffic source

nsrCodes:
Adds a persistent `hasConnectedApp` key to the global slice.

The modal is shown once the proxy has started.

https://github.com/requestly/requestly/assets/57226514/2bc9cda6-b67e-4e12-9434-080c989bec0d

### DIFF
--- a/app/src/store/index.js
+++ b/app/src/store/index.js
@@ -40,6 +40,7 @@ const globalReducer = getReducerWithLocalStorageSync("root", globalSlice.reducer
   "appTheme",
   // "rules",
   // "desktopSpecificDetails.appsList",
+  "hasConnectedApp",
   "userPersona",
   "country",
   "mobileDebugger",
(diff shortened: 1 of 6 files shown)


nsrCodes:
If we don't add the `isProxyRunning` check (which is what delays the opening of the modal, the user will be shown an empty modal like this 


https://github.com/requestly/requestly/assets/57226514/08bf3764-4474-4ed7-b8a1-3cb164a3b7d3

app/src/store/index.js
```diff
@@ -40,6 +40,7 @@ const globalReducer = getReducerWithLocalStorageSync("root", globalSlice.reducer
   "appTheme",
   // "rules",
   // "desktopSpecificDetails.appsList",
+  "hasConnectedApp",
```
wrongsahil, line 43: Why do we need to create another top level key here?  This can be added to `misc.persist`

wrongsahil (changes requested):

wrongsahil (approved):

PR #711, opened 2023-06-12
fix: modifying user was also overriding the rule's createdBy property

nsrCodes:


app/src/components/landing/ruleTemplates/RulePreviewModal/index.js
```diff
@@ -35,15 +35,15 @@ const RulePreviewModal = ({ rule, isOpen, toggle }) => {
         return;
       }
     }
-    const lastModifiedBy = Date.now();
+    const lastModified = Date.now();
     const modificationDate = Date.now();
```
nafees87n, line 39: `lastModified` and `modificationDate` are same thing. Can you please check why are we using them?

app/src/components/features/rules/RulesListContainer/RulesTable/index.js
```diff
@@ -920,7 +920,7 @@ const RulesTable = ({
         if (record.objectType && record.objectType === "group") {
           return null;
         }
-        const uid = record.createdBy ? record.createdBy : record.lastModifiedBy;
+        const uid = record.createdBy ?? null;
```
nafees87n, line 923: If `createdBy` is not defined then should we show ananymous?

nafees87n (changes requested):

app/src/components/features/rules/RulesListContainer/RulesTable/index.js
nsrCodes: yes that is what happens, when the uid here is null/undefined ([src](https://github.com/requestly/requestly/blob/master/app/src/components/common/UserIcon/index.tsx#L13))

The userIcon is shown as anonymous

app/src/components/landing/ruleTemplates/RulePreviewModal/index.js
nsrCodes: IIRC `modificationDate` was always part of the rule spec, but `lastModified` was added for entities that were introduced to workspaces. Keeping both here seems necessary. (to support both legacy code + new functionalities around team workspaces)
Although doesn't look relevant in the context of rule templates here....

wrongsahil: We can remove it as it is not consumed anywhere. I see only `modificationDate` getting consumed

nafees87n (approved):

PR #738, opened 2023-06-15
fix: expect query params from network logs

nsrCodes:
so far, query params were not being sent to the UI.
This is a supporting PR for the core fix inside the proxy - https://github.com/requestly/requestly-proxy/pull/20

### DIFF
--- a/app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficExporter/harLogs/converter.ts
+++ b/app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficExporter/harLogs/converter.ts
@@ -20,7 +20,7 @@ const convertLogToHarEntry = (log: Log) => {
     httpVersion: "HTTP/1.1",
     cookies: [],
     method: log.request.method,
-    queryString: [],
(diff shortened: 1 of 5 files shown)


app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficExporter/harLogs/converter.ts
```diff
@@ -83,10 +83,15 @@ export function createLogsHar(logs: Log[]) {
...
     const requestHeaders: Record<string, string> = {};
+    const requestQueryParams: Record<string, string> = {};
     entry.request.headers.forEach((headerObj) => {
       requestHeaders[headerObj.name] = headerObj.value;
     });
 
+    entry.request.queryString.forEach((queryObject) => {
+      requestQueryParams[queryObject.name] = queryObject.value;
+    });
+
```
wrongsahil, line 94: Can we use the queryString map from har directly instead of converting them into key: value pairs?  Don't see any reason we need to do this specifically

wrongsahil (changes requested):

wrongsahil (approved):

PR #761, opened 2023-06-21
RQ-493: update activate extension message

rohanmathur91:
Closes issue: 

## 📜 Summary of changes:
Screenshot 2023-06-21 at 11 56 40 AM

## ✅ Checklist:

- [ ] Verified UI in browser.


app/src/components/misc/ExtensionDeactivationMessage/index.js
```diff
@@ -1,28 +1,26 @@
...
+          <WarningOutlined /> Extension is paused
+        </h1>
+        <ol start="0">
+          <li>Please resume the extension by following below steps:</li>
+          <li>Open extension popup by clicking on Requestly icon on the browser's toolbar.</li>
+          <li>Hover over the dropdown menu on the popup that says "Requestly paused".</li>
+          <li>Click on "Resume requestly" to resume Requestly again.</li>
+        </ol>
+
+        <img width={628} height={235} src={resumeRequestly} alt="resume requestly" />
```
RuntimeTerror10, line 21: Can remove inline width and height given here as they are already defined in the CSS selector?

Also giving just the width will do the work, image will automatically adjust its height to maintain the aspect ratio.

RuntimeTerror10 (approved):

app/src/components/misc/ExtensionDeactivationMessage/index.js
rohanmathur91: this width and height will work as a placeholder, till the time our image is loading, reducing layout shift (in this case there won't be much), the width and height given from CSS will be applied when the image is loaded ie it will override `width={628}` and `height={235}`

PR #765, opened 2023-06-21
[RQ-524]: fix security issues in session recording

wrongsahil:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/app/src/backend/sessionRecording/getRecording.ts
+++ b/app/src/backend/sessionRecording/getRecording.ts
@@ -12,23 +12,6 @@ const isOwner = (recording: SessionRecording, uid: string) => {
   return false;
 };
 
-const hasRecordingAccess = (
-  recording: SessionRecording,
-  uid: string, // usrId or ownerId
-  email: string
-): boolean => {
-  if (recording.visibility === "public") return true;
-
-  if (!uid) return false;
-
-  if (recording.visibility === "only-me" && uid === recording.ownerId) return true;
-
-  // TODO: We should ideally keep uids here instead of keeping emails.
-  if (recording.accessEmails.includes(email) || uid === recording.ownerId) return true;
-
-  return false;
-};
-
 const fetchRecordingEvents = async (filePath: string) => {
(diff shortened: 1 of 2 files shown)


app/src/views/features/sessions/SessionViewer/SavedSessionViewer.tsx
```diff
@@ -119,9 +119,16 @@ const SavedSessionViewer: React.FC = () => {
...
       .catch((err) => {
-        err.name === "PermissionDenied" && setShowPermissionError(true);
-
-        err.name === "NotFound" && setShowNotFoundError(true);
+        switch (err.name) {
+          case "PermissionDenied":
+            setShowPermissionError(true);
+            break;
+          case "NotFound":
+            setShowNotFoundError(true);
```
lazyvab, line 127: minor - Moving `NotFound` case one level up, you won't need separate case for `PermissionDenied`.

app/src/backend/sessionRecording/getRecording.ts
```diff
@@ -24,7 +24,7 @@ const hasRecordingAccess = (
   if (recording.visibility === "only-me" && uid === recording.ownerId) return true;
 
   // TODO: We should ideally keep uids here instead of keeping emails.
-  if (recording.accessEmails.includes(email)) return true;
+  if (recording.accessEmails.includes(email) || uid === recording.ownerId) return true;
```
lazyvab, line 27: Do we need it if Firebase security rules are already enforced? I would rather like to remove `hasRecordingAccess` function from client-side altogether.

wrongsahil: Only the case where the recording is a workspace recording and user is trying to access the recording url from a different workspace breaks when `hasRecordingAccess` is removed. This can't be avoided as according custom claims, user has access to recording and we don't know whether user is on the workspace with recording or not

lazyvab (approved):

Issue #770, opened 2023-06-23
Rule name should be an HTML link


echo-sg:
### Issue
When you are on the Rules list page, the name of rules appear as links. But you cannot open a rule in new tab using Cmd+click or right-click and Open Link in New Tab.
<img width="1440" alt="Screenshot 2023-06-23 at 11 08 45 AM" src="https://github.com/requestly/requestly/assets/56088056/6a6c0d56-3f99-4653-a6cd-afdfcccf30ca">

### Repro steps
- Open Requestly's [Web App](https://app.requestly.io/).
- Create a rule if there aren't any.
- Open the [rules list page](https://app.requestly.io/rules/my-rules).

### Related issue
#305

ncharanaraj:
Hello, @echo-sg I will add target='_blank' and href="/#" to the rules anchor tag. Can you assign me this task?

sachinjain024:
Hey @ncharanaraj Any luck here?

ncharanaraj:
Hello, @sachinjain024 I figured it out and I'm requesting for review now.

lazyvab:
@ncharanaraj I checked your pull request and it does not meet the expectation. Do you want to give another try?

ncharanaraj:
@lazyvab I'll try

ncharanaraj:
@lazyvab Path has been updated

waheedshams:
hi can i work on this?

echo-sg:
@waheedshams can you pick this issue?

wwzeng1:
Hey @echo-sg! I can take this one :)

Nischalj10:
Hey @echo-sg, if this is done, can you mark this as closed?

wrongsahil:
This is not yet completed. Reopening this

ncharanaraj:
Hello Team,

I apologize for the delay in my response. I have implemented the required changes, and I've tested them locally. Everything seems to be working perfectly, including the ability to open links in a new tab using the "cmd/ctrl + click" and the right-click context menu.

You can watch a video clip demonstrating this functionality here: [Video Link](https://clipchamp.com/watch/93vl0G7oFm5)

I would appreciate it if @wrongsahil and @sachinjain024  could review the changes. Once I receive their feedback, I'll proceed with creating a pull request.

Please let me know. 

Thank you!.

DoyeDesigns:
Hello @echo-sg , have this issue been solved? I’d like to work on it.

echo-sg:
[...]

Issue #305, opened 2023-02-20
Make rules links actual HTML links


Samiii-AP:
In the interface, please use actual HTML links (<a href>) to allow us to open it in a new tab.

Currently, in HTTP Rules, the rules are links but impossible to open them in new tabs. This is a problem when we want to check the coherence of their content for example.

malay5:
Could you please provide more description of what is required?

lazyvab:
@malay5 When you are on the Rules list page, the name of rules appear as links. But you cannot open a rule in new tab using `Cmd+click` or right-click and `Open Link in New Tab`. 

This is most likely because the link is not an HTML link `<a>` and an on-click listener is attached using JS.

wrongsahil:
Duplicate of #770

PR #797, opened 2023-07-01
fixed rules as html links #770

ncharanaraj:
Closes issue:  #770

## 📜 Summary of changes:

Added href="/" to the Link component and called e.preventDefault()  to prevent the default browser behaviour. 

## ✅ Checklist:

- [x] Verified UI in browser.

## 🔗 Other references: 

 #305


app/src/components/features/rules/RulesListContainer/RulesTable/index.js
```diff
@@ -612,7 +612,9 @@ const RulesTable = ({
               }}
             >
               <Link
+                href="/"
```
lazyvab, line 615: `href="/"` means link to the root path. If you open this link in new tab, it is going to load the root path of the app which is the Rules list page itself.

The expectation was to open the rule definition in the new tab - that is, it should open the same screen that opens up on normal clicking of the link.

ncharanaraj: Thanks for the feedback. I'll fix the issue.

ncharanaraj: Hey @lazyvab updated the href path from "/" to {`${PATHS.RULE_EDITOR.EDIT_RULE.ABSOLUTE}/${record.id}`}. I hope I met the expectation.

app/src/components/features/rules/RulesListContainer/RulesTable/index.js
```diff
@@ -612,7 +612,9 @@ const RulesTable = ({
               }}
             >
               <Link
+                href={`${PATHS.RULE_EDITOR.EDIT_RULE.ABSOLUTE}/${record.id}`}
                 onClick={(e) => {
+                  e.preventDefault();
                   handleRuleNameOnClick(e, record);
```
RuntimeTerror10, line 618: Hey @ncharanaraj 
The current implementation has some issues :
* Does not open in new tab using Ctrl/Cmd + click
* This rule table is also being used in shared lists page. In the rule table on shared lists page we open the rule definition in a modal instead of navigating to a different route (check the `handleRuleNameOnClick` function definition). If I **right-click** and **open the link in new tab**, this redirects to `/404`. We don't require a `href` in case of shared list rule table.

echo-sg:
@ncharanaraj any luck on the suggestions suggested by @RuntimeTerror10?

echo-sg:
@ncharanaraj any luck on the suggestions by @RuntimeTerror10?

ncharanaraj:
Hello Team,

I apologize for the delay in my response. I have implemented the required changes, and I've tested them locally. Everything seems to be working perfectly, including the ability to open links in a new tab using the "cmd/ctrl + click" and the right-click context menu.

You can watch a video clip demonstrating this functionality here: [Video Link](https://clipchamp.com/watch/93vl0G7oFm5)

I would appreciate it if @echo-sg , @lazyvab, and @RuntimeTerror10 could review the changes. Once I receive their feedback, I'll proceed with creating a pull request.

Please let me know if you have any questions or need further changes.

PR #799, opened 2023-07-01
fix: navigate in useEffect dependency adds multiple ipc handlers

nsrCodes:


wrongsahil (approved):
Nice catch. This must have causing other issues too

PR #811, opened 2023-07-07
chore: add link to how to export from charles

nsrCodes:
Screenshot 2023-07-07 at 2 45 00 PM


echo-sg:
We have this event which could be triggered on link here: `charles_settings_import_docs_clicked`.
Please add source(event prop) as `upload_screen`.

echo-sg (approved):

PR #834, opened 2023-07-13
RQ-635 enable Logger in web app from devtools console 

RuntimeTerror10:
Closes issue: 

## 📜 Summary of changes:
* To enable Logger using query param add `debug` query param in URL
* To enable Logger from devtools console write **(only in web app)** :
`window.RQ.components.logger.enabled = true`

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/common/logger.js
+++ b/common/logger.js
@@ -24,10 +24,10 @@ if (urlParams.has("enableLogger")) {
   Logger.enabled = true;
 }
 
+window.RQ = window.RQ || {};
+window.RQ.components = window.RQ.components || {};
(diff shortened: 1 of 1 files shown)


nsrCodes (approved):

RuntimeTerror10 (commented):
Enabling logger for desktop app from devtools console dose not work for now as in web app`window.RQ` object is initialized from `/common` whereas in desktop app it is being initialized in `preloader.js`.

Was able to add the logger property in `window.RQ.components` but the logger does not get enable it because the render window does not have access make any changes `window` object.

PR #856, opened 2023-07-18
feat: Render duration of request in traffic table

nsrCodes:
Renamed current time column from `Time` to `Start Time`
Added a `Time` section that lists the duration of the request. Also adds prefixes like `ms`, `s`, `min`, etc.

Created a utility function for the same


nsrCodes:
Related proxy changes - https://github.com/requestly/requestly-proxy/pull/23

PR #873, opened 2023-07-24
[RQ-674] refactor: use source URL common component for session settings 

rohanmathur91:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/app/src/components/common/SourceUrl/index.tsx
+++ b/app/src/components/common/SourceUrl/index.tsx
@@ -1,16 +1,19 @@
 import React from "react";
 import { Col, Select, Input } from "antd";
 import { capitalize } from "lodash";
-import { Source } from "./types";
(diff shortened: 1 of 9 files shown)


app/src/components/common/SourceUrl/index.tsx
```diff
@@ -48,13 +51,16 @@ export const SourceConditionInput: React.FC<SourceProps> = ({ source, onSourceCh
         </Select>
       </Col>
       <Input
+        autoFocus
```
RuntimeTerror10, line 54: It is not necessary for the input in this component to be by default focussed for other instances we use or may use this component in  (eg. test url & rule editors ).
Can we make it auto focus by passing a prop to this component?

RuntimeTerror10 (approved):

PR #902, opened 2023-08-05
[RQ-702] fix: session viewer lagging when changing start time offset

nafees87n:
Closes issue:  RQ-702

## 📜 Summary of changes: 
The player would play from the offset time only on the page load and not when start offset is changed while playing the session.

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
+++ b/app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
@@ -103,7 +103,10 @@ const SessionDetails: React.FC = () => {
     }
 
     player.goto(startTimeOffset * 1000, true);
-  }, [player, startTimeOffset]);
+    // player should start playing from the start time offset only on the
+    // first load and not when the user changes time offset.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [player]);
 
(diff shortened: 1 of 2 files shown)


app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
```diff
@@ -103,7 +103,10 @@ const SessionDetails: React.FC = () => {
     }
 
     player.goto(startTimeOffset * 1000, true);
-  }, [player, startTimeOffset]);
+    // player should start playing from the start time offset only on the
+    // first load and not when the user changes time offset.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [player]);
```
lazyvab, line 109: Consider a case where player was at 5sec and I selected offset to 20sec. I would like to see the frame at 20sec. 
In case we don't want this, skipping dependency in `useEffect` is not the right way. We can add another state to component to keep track the player was initialized. 

Anyway, I think performance degradation is because of another issue. I see in `SessionPropertiesPanel` that we don't update redux store with debounce, so it triggers render cycle on every character stroke in the offset input.

lazyvab (approved):

PR #905, opened 2023-08-05
fix: add new attribute to track number of recordings created by user

nsrCodes:
adds the `num_sessions_saved_online` and `num_sessions_saved_offline` attribute 
does not change `num_sessions`

updates the onboarding flows to use these new attributes


app/src/views/features/sessions/SessionViewer/SavedSessionViewer.tsx
```diff
@@ -111,11 +111,16 @@ const SavedSessionViewer: React.FC = () => {
     });
   }, [id, eventsFilePath, navigateToList]);
 
+  const hasUserCreatedSessions = useMemo(
+    () => userAttributes?.num_sessions_saved_online > 0 || userAttributes?.num_sessions_saved_offline > 0,
+    [userAttributes?.num_sessions_saved_online, userAttributes?.num_sessions_saved_offline]
+  );
+
```
RuntimeTerror10, line 118: Existing users who have seen the setting banner may see it once again. Can we also add a check for `num_sessions` here.

RuntimeTerror10 (approved):

PR #933, opened 2023-08-11
[RQ-712] fix: current time greater than the video duration

nafees87n:
Closes issue:  RQ-712

## 📜 Summary of changes:

- `rrweb` listeners do not work when the focus is changed to another tab. Hence, the player keeps on playing.
- For fixing, this PR pauses the player `onBlur`.

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
+++ b/app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
@@ -129,6 +129,13 @@ const SessionDetails: React.FC<SessionDetailsProps> = ({ isInsideIframe = false 
       return;
     }
 
+    // no rrweb listener on the player works when focus is shifted from the tab
+    // The player keeps playing even when the tab is not in focus.
+    // so we add a listener on the window to pause the player when the tab is blurred
+    window.addEventListener("blur", () => {
+      player.pause();
+    });
+
     // player should start playing from the start time offset only on the
     // first load and not when the user changes time offset.
     player.goto(offsetTimeRef.current * 1000, true);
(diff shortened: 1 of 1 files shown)


app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
```diff
@@ -129,6 +129,13 @@ const SessionDetails: React.FC<SessionDetailsProps> = ({ isInsideIframe = false
       return;
     }
 
+    // no rrweb listener on the player works when focus is shifted from the tab
+    // The player keeps playing even when the tab is not in focus.
+    // so we add a listener on the window to pause the player when the tab is blurred
+    window.addEventListener("blur", () => {
```
lazyvab, line 135: Is it possible that on component remount, we might attach multiple such listeners on `window`, as we are not calling `removeEventListener` here?

nafees87n: Yes. It will happen. Removing listener on unmount.

app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
```diff
@@ -111,10 +115,11 @@ const SessionDetails: React.FC<SessionDetailsProps> = ({ isInsideIframe = false
   // destroy player on unmount
   useEffect(
     () => () => {
+      window.removeEventListener("blur", onBlurListener);
       // @ts-ignore
       player?.$destroy();
     },
-    [player]
+    [onBlurListener, player]
   );
```
lazyvab, line 123: You probably don't need to memoize the function. JS closure would have worked.
```js
useEffect(() => {
  const pauseVideo = () => {
    player?.pause();
  });

  window.addEventListener("blur", pauseVideo);

  return () => {
    // @ts-ignore
    player?.$destroy();
    window.removeEventListener("blur", pauseVideo);
  };
}, [player]);
```

lazyvab (approved):

PR #947, opened 2023-08-14
[RQ-759] feat: spacebar to play/pause the video

nafees87n:
Closes issue:  RQ-759

## 📜 Summary of changes:

Spacebar toggles the play/pause state of session replay player.

## ✅ Checklist:

- [ ] Verified UI in browser.


app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
```diff
@@ -142,6 +142,20 @@ const SessionDetails: React.FC<SessionDetailsProps> = ({ isInsideIframe = false
...
 
+  useEffect(() => {
+    const togglePlay = (e: KeyboardEvent) => {
+      e.preventDefault();
+      if (e.code === "Space") {
+        player?.toggle();
+      }
+    };
+
+    document.addEventListener("keydown", togglePlay);
```
rohanmathur91, line 153: This will listen to all the `keydown` events and will prevent the default behavior of the keys, so we won't be able to type. Let's have an another check of where this event has occurred along with `e.code === "Space"`

```diff
- e.preventDefault();
- if (e.code === "Space") {
+ if (e.code === "Space" && e.target === document.body) {
+   e.preventDefault();
+   player?.toggle();
+ }
```

rohanmathur91 (changes requested):

app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
lazyvab: @rohanmathur91 Should we rather check for `event.currentTarget` instead of `event.target`, so that if event happens on player video, it will bubble up to the body.

rohanmathur91: @lazyvab 
If we use `event.currentTarget` then we will lose the context of where the event was occurred, since any keydown event happening on input will also bubble up to document, so every time listener will get the `document` as the `currentTarget` but event might be triggered on body for play/pause or in an input for description so we can't distinguish between them.

rohanmathur91 (approved):

app/src/views/features/sessions/SessionViewer/SessionDetails.tsx
lazyvab: @rohanmathur91 That makes sense. Unlike mouse click event, key press event happens on document.body if no element is focused.

PR #1002, opened 2023-08-29
[RQ-808] : Shared list sharing 

RuntimeTerror10:
Closes issue: 

## 📜 Summary of changes:

https://github.com/requestly/requestly/assets/53986600/c643068a-4798-4483-9670-4e6001960acd

## ✅ Checklist:

- [ ] Make sure linting and unit tests pass.
- [ ] No install/build warnings introduced.
- [ ] Verified UI in browser.
- [ ] For UI changes, added/updated analytics events (if applicable).
- [ ] For changes in extension's code, manually tested in Chrome and Firefox.
- [ ] For changes in extension's code, both MV2 and MV3 are covered.
- [ ] Added/updated unit tests for this change.
- [ ] Raised pull request to update corresponding documentation (if already exists).

PR #1011, opened 2023-09-04
[RQ-809]: add download rules option in sharing modal

rohanmathur91:
Closes issue: 

## 📜 Summary of changes:
Screenshot 2023-09-04 at 2 36 26 PM

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/app/src/components/common/SharingModal/DownloadRules/index.tsx
+++ b/app/src/components/common/SharingModal/DownloadRules/index.tsx
@@ -1,0 +1,81 @@
+import React, { useCallback, useEffect, useMemo, useState } from "react";
+import { useDispatch, useSelector } from "react-redux";
+import { actions } from "store";
+import { Button } from "antd";
+import fileDownload from "js-file-download";
+import { getAllRules, getAppMode, getGroupwiseRulesToPopulate } from "store/selectors";
+import { prepareContentToExport } from "components/features/rules/ExportRulesModal/actions";
+import { trackRQLastActivity } from "utils/AnalyticsUtils";
+import { Rule } from "types";
+import { trackRulesExportedEvent } from "modules/analytics/events/common/rules";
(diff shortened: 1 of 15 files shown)


app/src/components/common/SharingModal/DownloadRules/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+      setRulesToDownload(null);
+    },
+    [fileName, rulesToDownload]
+  );
+
+  useEffect(() => {
+    if (!rulesToDownload) {
+      prepareContentToExport(appMode, selectedRules, groupwiseRulesToPopulate).then((result) => {
+        setRulesToDownload(result);
+        dispatch(actions.clearSelectedRules());
```
RuntimeTerror10, line 57: We should not clear rule selections immediately after opening modal (In case of false selections user may want to update the selections). We can clear this after any successful operations like after downloading for export, after sharedlist creation and same for workspace.

app/src/utils/DateTimeUtils.js
```diff
@@ -37,6 +37,16 @@ export const dateStringFromTimestamp = (timestamp) => {
...
+/** Returns date in DD_MM_YYYY format */
+export const getTodaysDateInFormat = () => {
+  const date = new Date();
+  const day = String(date.getDate()).padStart(2, "0");
+  const month = String(date.getMonth() + 1).padStart(2, "0");
+  const year = date.getFullYear();
+
+  return `${day}_${month}_${year}`;
+};
+
```
RuntimeTerror10, line 49: Instead of creating a new util, Can make use of `epochToDateAndTimeString`, pass it `new Date()` as arg and extract the date from the returned value. wdyt?
[reference](https://github.com/requestly/requestly/pull/1002/files#diff-78dd56e2bc55dd346199b77ba9fe18de59585ba424c50ef40eb76dfd7ea112dcR187)

PR #1081, opened 2023-09-25
[RQ-897]: pin extension popup

rohanmathur91:
## 📜 Summary of changes:
Pin extension popup

https://github.com/requestly/requestly/assets/61556757/01a07e59-8c8c-40a6-a26d-ad9bae8ef7b8

## ✅ Checklist:

- [ ] Verified UI in browser.


app/src/modules/analytics/events/common/onboarding/pinExtensionPopup.ts
```diff
@@ -0,0 +1,24 @@
+import { trackEvent } from "modules/analytics";
+import { ONBOARDING } from "../constants";
+
+type PinExtensionExpandedAction =
+  | "activate_deactivate"
+  | "what_rules_executed"
+  | "record_and_replay"
+  | "pause_extension";
+
+type PinExtensionPopupClosed = "close" | "already_pinned";
```
RuntimeTerror10, line 10: These types are defined at more than one place, instead export these from a separate `types` file.

app/src/components/common/PinExtensionPopup/hooks/usePinExtensionPopup.ts
```diff
@@ -0,0 +1,20 @@
...
+  const rulesCount = useSelector(getUserRulesCount);
+
+  const [isPinExtensionPopupActive, setIsPinExtensionPopupActive] = useState<boolean>(
+    JSON.parse(window.localStorage.getItem(IS_PIN_EXTENSION_POPUP_ACTIVE)) ?? rulesCount > 0
+  );
+
+  const closePinExtensionPopup = useCallback(() => {
+    setIsPinExtensionPopupActive(false);
+    window.localStorage.setItem(IS_PIN_EXTENSION_POPUP_ACTIVE, "false");
+  }, []);
```
RuntimeTerror10, line 17: Why are we using `window.localStorage` instead of redux store?

app/src/components/common/PinExtensionPopup/PinExtensionPopup.css
```diff
@@ -0,0 +1,157 @@
+.pin-extension-popup-container .rq-modal.ant-modal {
+  width: 373px;
+  min-width: 370px !important;
+  margin: 0px;
+  top: 16px;
+  left: calc(100% - 373px - 16px); /* whole screen - container width - spacing */
+  transform-origin: top right !important;
+}
```
RuntimeTerror10, line 8: Give a `className` to RQModal and add styles to it, instead of overwriting

RuntimeTerror10: We can remove `width` from here as the same is given as `width` prop to RQModal. Also why do we need` min-width` here?

RuntimeTerror10 (changes requested):

app/src/components/common/PinExtensionPopup/hooks/usePinExtensionPopup.ts
rohanmathur91: I think we should not put everything into redux store just for using localStorage via store.

app/src/components/common/PinExtensionPopup/PinExtensionPopup.css
rohanmathur91: This `min-width` is overriding the default RQModals min-width which is 550px

RuntimeTerror10 (approved):

PR #1109, opened 2023-10-03
fix: correctly populate rules metadata

nsrCodes:


wrongsahil:
Closing this due to inactivity. We can reopen this if when we decide to implement this

PR #1140, opened 2023-10-09
chore: add import-from-charles on a separate route

nsrCodes:
https://github.com/requestly/requestly/assets/57226514/352f0722-3c3a-4da9-be5a-9da91426a010


rohanmathur91 (commented):
- Why do we need separate route for Charles import?
- Once user is on rules page there is no touch point for charles import is it intentional ?

rohanmathur91:
UI doesn't look good, can we once confirm this with @wrongsahil 

Screenshot 2023-10-11 at 11 31 34 AM

rohanmathur91 (approved):

PR #1148, opened 2023-10-10
chore: add event param executor_is_creator

nsrCodes:
- Starts sending `rule_creator` in `rule_executed` event
- removes the old implementation for sending `rule_executed` events (adding unnecessary noise)
- adds event param `executor_is_creator`

### DIFF
--- a/app/src/modules/analytics/events/extension.ts
+++ b/app/src/modules/analytics/events/extension.ts
@@ -26,9 +26,9 @@ const sendEventsBatch = (eventBatch: EventBatch): void => {
 
       if (ruleCreator && currentUserId) {
         const isExecutorCreator = currentUserId === ruleCreator;
-        delete event.eventParams["rule_creator"];
         event.eventParams["is_executor_creator"] = isExecutorCreator;
       }
+      delete event.eventParams["rule_creator"];
     }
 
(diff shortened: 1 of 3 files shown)


app/src/modules/analytics/events/extension.ts
```diff
@@ -17,6 +18,19 @@ interface EventBatch {
 const sendEventsBatch = (eventBatch: EventBatch): void => {
   eventBatch.events.forEach((event) => {
     const eventConfig = { time: event.eventTs };
+
+    /* ADDING EXTRA INFO FOR RULE EXECUTION EVENTS */
+    if (event.eventName === RULES.RULE_EXECUTED) {
+      const currentUserId = window.uid;
+      const ruleCreator = event.eventParams?.["rule_creator"];
+
+      if (ruleCreator && currentUserId) {
```
nafees87n, line 27: If `ruleCreator` or `currentUserId` is not present then `eventParams` will contain `rule_creator`. Do we need to send `rule_creator`?

nsrCodes: no we don't. Will move the deletion outside the `if` scope 👍🏻

nafees87n (approved):

PR #1165, opened 2023-10-13
chore: pass status code as args inside modify response with code

nsrCodes:
currently only for fetch, could not get this working for requests made with XHR.
extension port for changes in proxy - https://github.com/requestly/requestly-proxy/pull/28


wrongsahil:
Closing this due to inactivity. We can reopen this if when we decide to implement this

PR #1174, opened 2023-10-19
feat: domain based email suggestions for all email input modals

nsrCodes:


app/src/components/common/EmailInputWithDomainBasedSuggestions/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+const EmailInputWithDomainBasedSuggestions: React.FC<Props> = ({ onChange, transparentBackground = false }) => {
+  const user = useSelector(getUserAuthDetails);
+  const userEmail = user?.details?.profile?.email;
+
+  const [allSuggestions, setAllSuggestions] = useState<string[]>([]);
+
+  const getOrganizationUsers = useMemo(() => httpsCallable(getFunctions(), "users-getOrganizationUsers"), []);
+
+  useEffect(() => {
+    if (!isCompanyEmail(userEmail)) return;
```
RuntimeTerror10, line 24: We should also check if the user is a `verified` business domain user

app/src/components/common/EmailInputWithDomainBasedSuggestions/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+        border: "none",
+        colors: {
+          ...theme.colors,
+          primary: "#2d2d2d",
+          primary25: "#2b2b2b",
+          neutral0: "#2d2d2d",
+          neutral10: "#323337", // tag background color
+          neutral80: "whitesmoke", // tag text color
+          danger: "whitesmoke", // tag cancel icon color
+          dangerLight: "#323337", // tag cancel background color
```
RuntimeTerror10, line 74: Can we use our css variables here instead of hard-coded colors codes?

app/src/components/common/EmailInputWithDomainBasedSuggestions/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  }, [getOrganizationUsers, userEmail]);
+
+  const optionsFromSuggestions = useMemo(() => {
+    return allSuggestions.map((suggestion) => {
+      return {
+        label: suggestion,
+        value: suggestion,
+      };
+    });
+  }, [allSuggestions]);
```
RuntimeTerror10, line 40: Can we avoid this memoization? Instead set this mapped data to `allSuggestions` state variable after getting organization users

RuntimeTerror10 (changes requested):

app/src/components/common/EmailInputWithDomainBasedSuggestions/index.tsx
nafees87n: We already have a verified check in backend `users-getOrganizationUsers`

RuntimeTerror10 (approved):

PR #1193, opened 2023-10-27
[RQ-1033]chore: add paid feature nudge viewed event

rohanmathur91:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/app/src/modules/analytics/events/common/pricing/index.ts
+++ b/app/src/modules/analytics/events/common/pricing/index.ts
@@ -1,9 +1,25 @@
 import { trackEvent } from "modules/analytics";
 import { PRICING } from "../constants";
+import { FeatureLimitType } from "hooks/featureLimiter/types";
+import { RuleType } from "types";
 
 type ViewPricingPlansClickSource = "crown" | "my_profile" | "feature_limiter_banner" | "workspace_upgrade";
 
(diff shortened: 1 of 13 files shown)


app/src/modules/analytics/events/common/pricing/index.ts
```diff
@@ -1,9 +1,25 @@
...
+
+export type FeatureType = Lowercase<keyof typeof FeatureLimitType>;
+
+export type PaidFeatureNudgeViewedSource =
+  | "rule_dropdown"
+  | "rule_selection_screen"
+  | "share_button"
+  | "share_modal"
+  | Lowercase<keyof typeof RuleType>;
+
```
nafees87n, line 21: Shouldn't we use enum here? One benefit I can see is that while assingment we can use `PaidFeatureNudgeViewedSource.rule_dropdown`. WDYT?

nafees87n (approved):

app/src/modules/analytics/events/common/pricing/index.ts
rohanmathur91: For this we have to import the enum `PaidFeatureNudgeViewedSource` everywhere were we are showing premium icon which I feel was not needed.

Issue #1132, opened 2023-10-06
[RQ-650] bug: Dropdowns selection not working properly in source filter modal


wrongsahil:
Click area of dropdowns in filter modal of Rules is text only. It should be the whole dropdown item.

https://app.requestly.io/sessions/saved/jdQYlXxVSn72yZl2nm2B

sagarsoni7:
The dynamic operations provided by the dropdown are no longer supported with the current mv3 implementation.

PR #1330, opened 2024-01-05
RQ-650 Dropdown bug fixed

DhruvBachani:
Click area of dropdowns in filter modal of Rules is text only. Made it to be the whole dropdown item as mentioned in the issue.

Closes issue: https://github.com/requestly/requestly/issues/1132

## 📜 Summary of changes:

Moved the onclick property to the Menu Item instead of span to allow clicking on the item itself in the dropdown rather than only text.

## ✅ Checklist:

- [x] Verified UI in browser.

## 🧪 Test instructions:

- Create a new HTTP rule 
- Select filters 
- Select any option from the dropdown next to Page URL

## 🔗 Other references:

N/A


sagarsoni7:
Thank you so much for taking the time to contribute. Due to restrictions in mv3 we could no longer maintain the support for this feature.

PR #1324, opened 2024-01-05
Fix: syncing merges local records to the wrong sync target

nsrCodes:
Closes issue: 

## ✅ Checklist:

- [X] Make sure linting and unit tests pass.
- [X] No install/build warnings introduced.
- [X] Verified UI in browser.
- [ ] Added/updated unit tests for this change.
- [X] Raised pull request to update corresponding documentation (if already exists).

## 🧪 Test instructions:

Error used to happen when heavy operations happen across multiple simultaneous sessions on the same workspace
To make sure the bug occurs, switch to personal workspace at the end of the operation

## 🔗 Other references:
syncing diagram - https://excalidraw.com/#json=uqYbyi0s29VoRpJ67ALrA,R1fHpBIOiZ5DRfAKRJA9zw (error flow is in red)

PR #1366, opened 2024-01-25
Revert "Fix: syncing merges local records to the wrong sync target"

sagarsoni7:
Reverts requestly/requestly#1324


Kanishkrawatt (commented):
Lgtm

PR #1369, opened 2024-01-29
feat: elements with attributes in insert script rule

nsrCodes:
Core functionality to add attributes via the code in insert script rule
adding the optional `wrapperElement` attribute for script type code

### DIFF
--- a/browser-extension/mv2/src/client/js/scriptRuleHandler.js
+++ b/browser-extension/mv2/src/client/js/scriptRuleHandler.js
@@ -131,7 +131,17 @@ RQ.ScriptRuleHandler.includeJSScriptsInOrder = function (scripts, callback, inde
 RQ.ScriptRuleHandler.includeJS = function (script, callback) {
   if (!script.value) throw new Error("Script value is empty");
   if (script.type === RQ.SCRIPT_TYPES.URL) {
-    // todo: add attributes here
+    if (script.wrapperElement) {
+      // todo: add this to ruleBuilder and schema
+      const scriptBlock = RQ.ScriptRuleHandler.parseSpecificHTMLStringBlocks(script.wrapperElement, "script")[0];
+      if (scriptBlock) {
+        scriptBlock.attributes["src"] = script.value;
+        scriptBlock.innerText = "";
+        RQ.ScriptRuleHandler.embedBlocks("script", [scriptBlock]);
+        return;
+      }
+    }
+
     RQ.ClientUtils.addRemoteJS(script.value, callback);
     return;
   }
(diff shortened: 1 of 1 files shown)


browser-extension/mv2/src/client/js/scriptRuleHandler.js
```diff
@@ -128,24 +128,105 @@ RQ.ScriptRuleHandler.includeJSScriptsInOrder = function (scripts, callback, inde
 };
 
 RQ.ScriptRuleHandler.includeJS = function (script, callback) {
+  if (!script.value) throw new Error("Script value is empty");
   if (script.type === RQ.SCRIPT_TYPES.URL) {
+    if (script.wrapperElement) {
+      const scriptBlock = RQ.ScriptRuleHandler.parseSpecificHTMLStringBlocks(script.wrapperElement, "script")?.[0];
+      if (scriptBlock) {
```
nafees87n, line 135: Too many nested ifs. Can we convert script.type check into switch-case? Can we create another function which takes the required args and return the scriptBlock to be embedded?

browser-extension/mv2/src/client/js/scriptRuleHandler.js
```diff
@@ -128,24 +128,105 @@ RQ.ScriptRuleHandler.includeJSScriptsInOrder = function (scripts, callback, inde
...
+    };
+  });
+  return blockDetails;
+};
+
+RQ.ScriptRuleHandler.embedBlocks = function (htmlElementName, blockDetails) {
+  if (!htmlElementName || htmlElementName === "body") throw new Error("htmlElementName is empty");
+  blockDetails.forEach((block) => {
+    const htmlBlock = document.createElement(htmlElementName);
+    htmlBlock.className = RQ.ClientUtils.getScriptClassAttribute();
```
nafees87n, line 209: `htmlBlock.className` is never used. RQ specific class is never being added

browser-extension/mv2/src/client/js/scriptRuleHandler.js
```diff
@@ -128,24 +128,105 @@ RQ.ScriptRuleHandler.includeJSScriptsInOrder = function (scripts, callback, inde
...
+    RQ.ScriptRuleHandler.appendToDom(htmlBlock);
+  });
+};
+
+RQ.ScriptRuleHandler.appendToDom = function (htmlNode) {
+  const parent = document.head || document.documentElement;
+  try {
+    parent.appendChild(htmlNode);
+  } catch (error) {
+    console.error(error);
```
nafees87n, line 230: In case of error, shouldn't we log something user friendly instead of the actual error?

nafees87n (changes requested):

browser-extension/mv2/src/client/js/scriptRuleHandler.js
```diff
@@ -128,24 +128,105 @@ RQ.ScriptRuleHandler.includeJSScriptsInOrder = function (scripts, callback, inde
...
+      if (scriptBlock) {
+        scriptBlock.attributes["src"] = script.value;
+        scriptBlock.innerText = "";
+        RQ.ScriptRuleHandler.embedBlocks("script", [scriptBlock]);
+        return;
+      }
+    }
+
     RQ.ClientUtils.addRemoteJS(script.value, callback);
     return;
```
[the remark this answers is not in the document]
nsrCodes: missed this, adding

browser-extension/mv2/src/client/js/scriptRuleHandler.js
```diff
@@ -128,24 +128,105 @@ RQ.ScriptRuleHandler.includeJSScriptsInOrder = function (scripts, callback, inde
 };
 
 RQ.ScriptRuleHandler.includeJS = function (script, callback) {
+  if (!script.value) throw new Error("Script value is empty");
```
sachinjain024, line 131: Instead of throwing an error, should we do `console.error` and add a more human-readable log like `Script value used in Requeslty Inject Script Rule is empty.

We should make sure, it doesn't break the webpage just that the rule isn't run.

nsrCodes:
superseded by - https://github.com/requestly/requestly/pull/1386

PR #1386, opened 2024-02-05
feat: Attributes support for insert script

nsrCodes:
## Schema changes
- Adds `attributes` attribute to the script object (so inside the schema it is rule > pair > script.attributes)
- `script.value` still only contains the raw code. 

#### During Editing
- only one html node
- only the correct html node
- no `innerText` for html of script rule of `url` type

#### On Save:
- if JS code has `document.onLoad` in any way, and loadtime is `AfterPageLoad`: updates load time and shows editor toast, rule is not saved tho
- if, even after all the ongoing validations, `script.value` contains html code, error toast is shown 

## New functionalities:
Code editor now have their own toasts

# New UI:
Screenshot 2024-02-05 at 5 47 07 AM
Screenshot 2024-02-05 at 2 32 37 AM


app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,103 @@ import { actions } from "../../../../../../../../store";
...
+export const transformAndValidateInputsforRule = async (ruleData) => {
+  switch (ruleData.ruleType) {
+    case GLOBAL_CONSTANTS.RULE_TYPES.SCRIPT: {
+      const isCompatibleWithAttributesForScriptRule = isFeatureCompatible(FEATURES.SCRIPT_RULE_HTML_BLOCK);
+
+      if (!isCompatibleWithAttributesForScriptRule) {
+        return ruleData;
+      }
+
+      const ruleDataCopy = _.cloneDeep(ruleData);
```
sachinjain024, line 40: Why do we need a deep copy of rule data? Can you also add a comment here?

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,103 @@ import { actions } from "../../../../../../../../store";
...
+ * }
+ */
+export const transformAndValidateInputsforRule = async (ruleData) => {
+  switch (ruleData.ruleType) {
+    case GLOBAL_CONSTANTS.RULE_TYPES.SCRIPT: {
+      const isCompatibleWithAttributesForScriptRule = isFeatureCompatible(FEATURES.SCRIPT_RULE_HTML_BLOCK);
+
+      if (!isCompatibleWithAttributesForScriptRule) {
+        return ruleData;
+      }
```
sachinjain024, line 38: This should be moved into a separate function like `checkScriptRuleVersionCompatibility()`

```
function checkcheckScriptRuleVersionCompatibility() {
    // Check for Attributes support added in Feb 2024
    const isCompatibleForAttribues = isFeatureCompatible(FEATURES.SCRIPT_RULE_ATTRIBUTES_SUPPORT);
    
    if (!isCompatibleForAttribues) return false;

    // We can similarly check other compatible things in future

}
```

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,103 @@ import { actions } from "../../../../../../../../store";
...
+            };
+          } else if (script.type === GLOBAL_CONSTANTS.SCRIPT_TYPES.URL) {
+            const res = await validateStringForSpecificHTMLNode(script.wrapperElement, htmlNodeName);
+            if (!res.isValid) {
+              return {
+                error: res.validationError,
+                success: false,
+                message: res.errorMessage,
+                pairId: pair.id,
+              };
```
sachinjain024, line 79: I think we should break down this code into smaller functions. It will also help in writing UTs.

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
nsrCodes: RuleData is explicitly read-only (reference to value in store). Updating it directly is possible but leads to very messy code

Added a comment

sachinjain024: @nsrCodes I am still not sure you really need to deep clone the ruleData

Looking at this code below 👇 You are again creating a new ruleObject with new rulePairs

image

Just sharing my understanding of `...` operator 

image

sachinjain024: And, Please do not resolve any conversation just after your comment, let me resolve the conversation otherwise I will miss the responses.

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,104 @@ import { actions } from "../../../../../../../../store";
...
+import {
+  extractInnerTextAndAttributesFromHTMLString,
+  getHTMLNodeName,
+  validateStringForSpecificHTMLNode,
+} from "./insertScriptValidators";
+import { isFeatureCompatible } from "utils/CompatibilityUtils";
+import FEATURES from "config/constants/sub/features";
+
+/**
+ * Convert user inputs into rule schema
```
sachinjain024, line 18: This needs a better comment. This doc string doesn't explain anything.

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,104 @@ import { actions } from "../../../../../../../../store";
...
+      const newPairs = [];
+
+      for (const pair of ruleDataCopy.pairs) {
+        const parsedPair = pair;
+        const newScripts = [];
+
+        for (const script of pair.scripts) {
+          let newScript = script;
+
+          const htmlNodeName = getHTMLNodeName(script.type, script.codeType);
```
sachinjain024, line 51: Can we rename `htmlNodeName` to `codeHTMLTagName`?

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,104 @@ import { actions } from "../../../../../../../../store";
...
+              script.value,
+              htmlNodeName
+            );
+            newScript = {
+              ...script,
+              attributes,
+              value: code,
+            };
+          } else if (script.type === GLOBAL_CONSTANTS.SCRIPT_TYPES.URL) {
+            const res = await validateStringForSpecificHTMLNode(script.wrapperElement, htmlNodeName);
```
sachinjain024, line 73: Do we not have wrapper element in the custom code block?

sachinjain024: My suggestion is to keep the schema consistent in both the options from the UI side when a rule is getting created.

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,104 @@ import { actions } from "../../../../../../../../store";
...
+            if (!res.isValid) {
```
sachinjain024, line 74: Similarly please break down this into separate function

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,104 @@ import { actions } from "../../../../../../../../store";
...
+            newScript = {
+              ...script,
+              attributes,
+            };
+          }
+          delete newScript["wrapperElement"]; // this was used to temporarily store the raw code
+          newScripts.push(newScript);
+        }
+        parsedPair.scripts = newScripts;
+        newPairs.push(parsedPair);
```
sachinjain024, line 92: One approach to reformat this code as

1. Instead of deep cloning the entire rule data, you just loop on the rulePairs
2. And just update the script object inside the rulePair upon transformation

It won't make code messy but should simplify the code. You will not have to use so many copies, like newScript, parsedRulePair etc.

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,104 @@ import { actions } from "../../../../../../../../store";
...
+      }
+      const newRuleData = {
+        ...ruleData,
+        pairs: newPairs,
+      };
```
sachinjain024, line 97: This is so confusing! Once you have created copy of ruleData then why are you using the default ruleData to create the newRuleData at the bottom of the function.

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -217,14 +310,62 @@ export const validateRule = (rule, dispatch, appMode) => {
...
+                message: `Please enter a valid script URL`,
+                error: "invalid script url",
+              };
+            }
+            if (!isValidUrl(script.value)) {
+              output = {
+                result: false,
+                message: `Please enter a valid script URL`,
+                error: "invalid script url",
+              };
```
sachinjain024, line 367: If the error is same in both the cases then we can merge them with `||` operator. No?

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
```diff
@@ -0,0 +1,314 @@
...
+import { CONSTANTS as GLOBAL_CONSTANTS } from "@requestly/requestly-core";
+
+/* LOGICAL VALIDATORS - currently not being fully applied */
+export const scriptLogicalErrors = {
+  DOM_LOAD_EVENT_LISTENER_AFTER_PAGE_LOAD: "dom_load_event_listener_after_page_load",
+  CONTAINS_HTML_CODE: "contains_html_code",
+};
+
+export function checkForLogicalErrorsInCode(code, script) {
+  if (doesStringContainHTMLSnippet(code)) {
```
sachinjain024, line 13: `checkForNonJsOrCssCode` because that's the intent.

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
```diff
@@ -0,0 +1,314 @@
...
+    return {
+      isValid: false,
+      error: scriptLogicalErrors.CONTAINS_HTML_CODE,
```
sachinjain024, line 16: Similarly, update the error type `CONTAINS_NON_JS_CSS`

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
```diff
@@ -0,0 +1,314 @@
...
+      Array.from(doc.head.childNodes).some((node) => node.nodeType === 1) // special nodes like link and style end up in head
+    );
+  } catch (error) {
+    console.error("Error parsing code:", error);
+    return false;
+  }
+}
+
+/* HTML VALIDATORS */
+export const invalidHTMLError = {
```
sachinjain024, line 82: This should be in CAPS `HTML_ERRORS.COULD_NOT_PARSE`

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
```diff
@@ -0,0 +1,314 @@
...
+  };
+}
+
+/**
+ * Check if the given string contains HTML nodes
+ * @param {string} str string containing HTML snippet
+ * @param {Document} doc Document created from the HTML string
+ * @returns {boolean}
+ */
+function isHTMLString(str, doc) {
```
sachinjain024, line 211: What is the difference between `isHTMLString` and `doesContainHTMLString` function that you defined above?

https://github.com/requestly/requestly/pull/1386/files#diff-9ba7447af01ba32e558b219199e9e0247e03e17cc3dfdb9e97ed91ccd1e170d9R67

I see the code is being duplicated. Can we avoid it?

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
```diff
@@ -0,0 +1,314 @@
...
+  );
+}
+
+/**
+ * Checks if the document has nodes other than the given nodeName
+ * @param {Document} doc document to be checked. This is expected to be created from some HTML snippet string
+ * @param {string} nodeName name of the node that is expected to be present in the document
+ * @returns {boolean}
+ */
+function doesDocumentHaveOtherNodes(doc, nodeName) {
```
sachinjain024, line 236: checkDocumentForAnyOtherNode

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
```diff
@@ -0,0 +1,314 @@
...
+ */
+function isHTMLString(str, doc) {
+  // todo: rename
+  const defaultResponse = true;
+  if (!str) return defaultResponse;
+
+  // check if body and head is just empty
+  if (doc.body.childNodes.length === 0 && doc.head.childNodes.length === 0) {
+    /* since string exists, but doc is empty, this implies there was an error parsing it into the document */
+    console.log("6");
```
sachinjain024, line 219: Remove this console log

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
```diff
@@ -0,0 +1,314 @@
...
+    err: "Unexpected Error while parsing code",
+  };
+}
+
+export function getHTMLNodeName(scriptType, codeType) {
+  let htmlNode = "";
+  if (codeType === GLOBAL_CONSTANTS.SCRIPT_CODE_TYPES.JS) {
+    // else CSS
+    htmlNode = "script";
+  } else if (scriptType === GLOBAL_CONSTANTS.SCRIPT_TYPES.URL) {
```
sachinjain024, line 292: We should still wrap this under codeType check

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
```diff
@@ -0,0 +1,314 @@
...
+    innerText: undefined,
+    attributes: undefined,
+    err: "Unexpected Error while parsing code",
+  };
+}
+
+export function getHTMLNodeName(scriptType, codeType) {
+  let htmlNode = "";
+  if (codeType === GLOBAL_CONSTANTS.SCRIPT_CODE_TYPES.JS) {
+    // else CSS
```
sachinjain024, line 290: This comment isn't placed correctly

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
```diff
@@ -0,0 +1,314 @@
...
+    htmlNode = "script";
+  } else if (scriptType === GLOBAL_CONSTANTS.SCRIPT_TYPES.URL) {
+    htmlNode = "link";
+  } else {
+    htmlNode = "style";
+  }
+  return htmlNode;
+}
+
+function extractDomNodeDetailsFromHTMLCodeString(htmlCodeString, nodeName) {
```
sachinjain024, line 300: extractDOMNodeDetails

app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/CustomScriptRow.css
```diff
@@ -17,3 +17,7 @@
   border-bottom-left-radius: var(--border-radius-sm);
   border-bottom-right-radius: var(--border-radius-sm);
 }
+
+.code-editor-small {
+  max-height: 72px;
```
sachinjain024, line 22: @rohanmathur91 @RuntimeTerror10 -- You can review how we put styles

app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.js
```diff
@@ -9,27 +9,113 @@ import { CONSTANTS as GLOBAL_CONSTANTS } from "@requestly/requestly-core";
...
   const [isCodeFormatted, setIsCodeFormatted] = useState(false);
 
-  const scriptEditorBoilerCode = useMemo(
-    () => ({
-      JS: 'console.log("Hello World");',
-      CSS: "body {\n\t background-color: #fff;\n }",
-    }),
-    []
-  );
+  const isCompatibleWithAttributes = isFeatureCompatible(FEATURES.SCRIPT_RULE_HTML_BLOCK);
```
sachinjain024, line 78: Please rename to FEATURES.SCRIPT_RULE.ATTRIBUTES_SUPPORT because we still don't support HTML block. It sounds confusing

app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.js
```diff
@@ -47,8 +133,66 @@ const CustomScriptRow = ({ rowIndex, pairIndex, isLastIndex, deleteScript, scrip
...
+            span={24}
+            style={{
+              alignItems: "center",
+              justifyContent: "center",
+            }}
+          >
+            <Col xl="12" span={24}>
+              <CodeEditor
+                id={pair.id}
+                height={75}
```
sachinjain024, line 168: @RuntimeTerror10 @rohanmathur91 Do we specify fixed values like this?

app/src/components/misc/CodeEditor/EditorToast/editorToast.scss
```diff
@@ -0,0 +1,55 @@
...
+  font-size: 14px;
+  font-weight: 500;
+  cursor: pointer;
+
+  background-color: transparent;
+  border: none;
+  box-shadow: none;
+  height: 100%;
+
+  align-self: center;
```
sachinjain024, line 38: @RuntimeTerror10 -- I am leaving all the CSS and UI code changes to you.

browser-extension/mv2/src/client/js/scriptRuleHandler.js
```diff
@@ -128,24 +128,98 @@ RQ.ScriptRuleHandler.includeJSScriptsInOrder = function (scripts, callback, inde
 };
 
 RQ.ScriptRuleHandler.includeJS = function (script, callback) {
+  if (!script.value) throw new Error("Script value is empty");
   if (script.type === RQ.SCRIPT_TYPES.URL) {
-    RQ.ClientUtils.addRemoteJS(script.value, callback);
+    if (script.attributes) {
+      RQ.ScriptRuleHandler.addRemoteJSWithAttributes(script.value, script.attributes, callback);
```
sachinjain024, line 134: We shouldn't create a separate function but instead just pass the attributes and the addRemoteJS function should check if there are attributes then add them otherwise not.

app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.js
```diff
@@ -47,8 +133,66 @@ const CustomScriptRow = ({ rowIndex, pairIndex, isLastIndex, deleteScript, scrip
...
+          })
+        );
+      };
+
+      const handleCodeFormattedFlag = () => {
+        setIsCodeFormatted(true);
+        setTimeout(() => {
+          setIsCodeFormatted(false);
+        }, 2000);
+      };
```
RuntimeTerror10, line 154: This function is duplicated inside `renderCodeEditor` also.

app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.js
```diff
@@ -47,8 +133,66 @@ const CustomScriptRow = ({ rowIndex, pairIndex, isLastIndex, deleteScript, scrip
   };
 
   const renderURLInput = () => {
+    const renderCodeEditorForScriptAttributes = () => {
```
RuntimeTerror10, line 136: Can be very confusing.
`renderCodeEditorForScriptAttributes` inside  `renderURLInput`.
Can we have a common code editor component which can be wrapper for `CodeEditor` pass props accordingly for `script` and `url` type code editor.
This will also remove code duplication in both editor render functions

app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.js
RuntimeTerror10: Yes, the code editor height is resizable by dragging so we need to specify a fixed minimum height for the code editor.

RuntimeTerror10 (changes requested):

app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.js
RuntimeTerror10: Yes, the code editor height is resizable by dragging so we need to specify a fixed minimum height for the code editor.

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
nsrCodes: changed `doesContainHTMLString` to `isJSOrCSS` cause as you mentioned, that was the intent there

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/insertScriptValidators.js
nsrCodes: renamed to `isJSOrCSS` instead

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
nsrCodes: doing this just for this check would go agains the current convention across the code where we just use `isFeatureCompatible(<feature>)` for all such similar checks

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
nsrCodes: This works for when we only need to change 1 level deep. 
Here the whole object immutable, down to every scrip object which is why had to deep clone

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
```diff
@@ -3,10 +3,104 @@ import { actions } from "../../../../../../../../store";
...
+            }
+            const { attributes, innerText: code } = extractInnerTextAndAttributesFromHTMLString(
+              script.value,
+              htmlNodeName
+            );
+            newScript = {
+              ...script,
+              attributes,
+              value: code,
+            };
```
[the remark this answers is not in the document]
nsrCodes: moving these checks to a separate function seemed like overkill since they were completely independent and also had to manipulate the `newRuleData` object in the parent scope. 

But I hope the rewrite justifies this logic staying here

I have renamed the functions according to the suggestions

app/src/components/features/rules/RuleBuilder/Header/ActionButtons/CreateRuleButton/actions/index.js
nsrCodes: but extracting that to a variable here does seem redundant so replaced that with the function call itself

browser-extension/mv2/src/client/js/scriptRuleHandler.js
nsrCodes: updated this in a separate PR to make it easy for @nafees87n to also review it - https://github.com/requestly/requestly/pull/1400

RuntimeTerror10 (approved):

PR #1407, opened 2024-02-11
add patch-package to install script

sagarsoni7:


nsrCodes (approved):
> This is currently needed for the patch inside session recording.

PR #1433, opened 2024-02-19
RQ-1323: Commercial license banner

RuntimeTerror10:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.


app/src/layouts/DashboardLayout/AppNotificationBanner/index.tsx
```diff
@@ -24,10 +29,44 @@ interface Banner {
```
nsrCodes, line 29: You also need to expand this interface accordingly.
Looking through the dummy testing banners and the following code, quickly visible suggestions are:
1. `cta` is now optional
2. `actions` does not exist right now
3. `badgeColor` is a new optional attribute

app/src/layouts/DashboardLayout/AppNotificationBanner/index.tsx
```diff
@@ -41,20 +80,35 @@ export const AppNotificationBanner = () => {
...
-            <ReactMarkdown children={banner?.text} skipHtml={false} rehypePlugins={[rehypeRaw]} />
+            <ReactMarkdown
+              children={renderBannerText(banner.id, banner.text)}
+              skipHtml={false}
+              // @ts-ignore
+              rehypePlugins={[rehypeRaw]}
+            />
+          </div>
+          <div className="app-banner-action-buttons">
+            {banner.actions?.map((action: BANNER_ACTIONS) => {
```
nsrCodes, line 102: `actions` does not exist on current banner interface

app/src/layouts/DashboardLayout/AppNotificationBanner/index.tsx
nsrCodes: If we are planning to work on improving this interface futher, can we rename `short_text` to `shortText`?

app/src/layouts/DashboardLayout/AppNotificationBanner/index.tsx
```diff
@@ -41,20 +80,35 @@ export const AppNotificationBanner = () => {
...
+              {banner.short_text}
+            </span>
+          )}
           <div className="app-banner-text">
-            {/* @ts-expect-error */}
-            <ReactMarkdown children={banner?.text} skipHtml={false} rehypePlugins={[rehypeRaw]} />
+            <ReactMarkdown
+              children={renderBannerText(banner.id, banner.text)}
+              skipHtml={false}
+              // @ts-ignore
```
nsrCodes, line 97: This does not seem necessary as we can easily mitigate the lint error by upgrading `react-markdown` to v9+
All the module upgrades will anyway be picked up soon, and this seems like an easy one to start with since the module is only being consumed here.

you can refer to [this changelog](https://github.com/remarkjs/react-markdown/releases/tag/9.0.0) (_although on a cursor reading, doesn't feel like we will have make any changes to the code to upgrade to the latest version_)

app/src/layouts/DashboardLayout/AppNotificationBanner/types.ts
```diff
@@ -0,0 +1,4 @@
+export enum BANNER_ACTIONS {
```
nsrCodes, line 1: do these need to be exported? will they be consumed elsewhere? if not, can we also make these part of the original file near the `Banner` interface

app/src/layouts/DashboardLayout/AppNotificationBanner/index.tsx
nsrCodes: Also document the relationship/dependency between `short_text` and `text` in some doc-string or comment, and any other known assumptions. Since will be the main source to understand how to structure the banner JSON

nsrCodes (changes requested):

PR #1469, opened 2024-02-29
RQ-1355: seed user properties

RuntimeTerror10:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/app/src/hooks/AuthHandler.js
+++ b/app/src/hooks/AuthHandler.js
@@ -40,7 +40,7 @@ const AuthHandler = (onComplete) => {
         window.uid = authData?.uid;
         localStorage.setItem("__rq_uid", authData?.uid);
 
-        // New user signing in from an org domain will be assigned a serial number attribute
+        /* To ensure that this attribute is assigned to only the users signing up for the first time */
         if (user?.metadata?.creationTime === user?.metadata?.lastSignInTime) {
           if (isCompanyEmail(user.email) && user.emailVerified && !userAttributes[TRACKING.ATTR.COMPANY_USER_SERIAL]) {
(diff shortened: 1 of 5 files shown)


app/src/hooks/AuthHandler.js
```diff
@@ -38,6 +39,15 @@ const AuthHandler = (onComplete) => {
         window.uid = authData?.uid;
         localStorage.setItem("__rq_uid", authData?.uid);
 
+        if (user.metadata.creationTime === user.metadata.lastSignInTime) {
```
nsrCodes, line 42: doesn't seem reliable, can we use `result.additionalUserInfo.isNewUser` instead?

nsrCodes: and have we tested if this is working as expected, maybe using console logs?

nsrCodes (changes requested):

app/src/hooks/AuthHandler.js
nsrCodes: @RuntimeTerror10 with [the updated changes](https://github.com/requestly/requestly/pull/1469/commits/af6d33ebc4c2639bbfff6ec59b83c07248725b9a) this check no longer seems required

app/src/hooks/AuthHandler.js
```diff
@@ -38,6 +40,15 @@ const AuthHandler = (onComplete) => {
...
```
nsrCodes, line 43: Also user will be null when logging out. so either add optional chaining or a null check to avoid `null.metadata`

nsrCodes (approved):

PR #1520, opened 2024-03-21
fix: don't show as unsaved changes for first render of script rule

nsrCodes:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.js
+++ b/app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.js
@@ -1,5 +1,5 @@
 /* eslint-disable default-case */
-import React, { useCallback, useEffect, useMemo, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.js
```diff
@@ -186,6 +187,12 @@ const CustomScriptRow = ({
 
   const handleEditorUpdate = useCallback(
     (value) => {
+      const shouldTriggerUnsavedChanges = !isFirstRender.current;
+
+      if (isFirstRender.current) {
+        isFirstRender.current = false;
+      }
+
```
RuntimeTerror10, line 195: Instead of creating a `ref` to check for first render, Can we pass an optional arg `triggerUnsavedChanges` which is by default `true` and in the function call inside useEffect we pass `false`. Then we don't need to add extra logic for first render in editor update function

RuntimeTerror10 (changes requested):

RuntimeTerror10 (approved):

PR #1534, opened 2024-03-29
feat: compress pair values of certain rule types

nsrCodes:
compresses the pair value for 
1. Response rule
2. Request rule
3. Script rule that are of code type

in transit (to and from firebase) and storage. 
Syncing still works on the complete (uncompressed) records

### DIFF
--- a/app/src/utils/Compression.js
+++ b/app/src/utils/Compression.js
@@ -42,7 +42,7 @@ function decompressRecord(record) {
         });
         break;
 
-      case "Script":
+      case RuleType.SCRIPT:
         record.pairs.forEach((pair, idx) => {
           pair.scripts.forEach((script, sidx) => {
             if (script.type === "code" && script.isCompressed) {
@@ -92,7 +92,7 @@ function compressRecord(record) {
         });
         break;
 
-      case "Script":
+      case RuleType.SCRIPT:
         record.pairs.forEach((pair, idx) => {
           pair.scripts.forEach((script, sidx) => {
             if (script.type === "code") {
--- a/app/src/utils/Compression.js
+++ b/app/src/utils/Compression.js
@@ -14,12 +14,6 @@ function compressString(uncompressedString) {
   return base64Compressed;
 }
 
-function shouldDecompressRecord(recordId) {
-  // trying to keep it as efficient as possible by working with ID instead of whole object
-  const recordType = recordId.split("_")[0];
-  return recordType && ["Response", "Request", "Script"].includes(recordType);
-}
-
 function decompressRecord(record) {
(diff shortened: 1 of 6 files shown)


sagarsoni7:
We’re first compressing the rule data and then encoding it to base64.
Wouldn’t encoding to base64 increase the size again?
If we are doing base64 encoding to convert it into text to set data in Realtime db, how about directly pushing the compressed outcome to Firestore DB instead? It accepts ByteString

For now, we can move forward with RealtimeDB but we still have the future scope to compress further by moving rules to FirestoreDB right? Of course Firestore has its own challenges we know

sagarsoni7 (approved):

app/src/utils/Compression.js
```diff
@@ -0,0 +1,118 @@
...
+          pair.scripts.forEach((script, sidx) => {
+            if (script.type === "code") {
+              compressedRecord.pairs[idx].scripts[sidx].value = compressString(script?.value ?? "");
+              compressedRecord.pairs[idx].scripts[sidx].isCompressed = true;
+            }
+          });
+        });
+        break;
+      default:
+        console.log("This rule type should not need compression");
```
wrongsahil, line 106: Remove log

app/src/utils/Compression.js
```diff
@@ -0,0 +1,118 @@
...
+        console.log("This rule type should not need decompression");
+    }
+  }
+  return decompressedRecord;
+}
+
+export function decompressRecords(recordsMap) {
+  const decompressedRecordsMap = {};
+  for (const recordId in recordsMap) {
+    if (shouldDecompressRecord(recordId)) {
```
wrongsahil, line 66: This looks redundant as this check is already present in decompressRecords.

app/src/utils/Compression.js
```diff
@@ -0,0 +1,118 @@
...
+
+function shouldDecompressRecord(recordId) {
+  // trying to keep it as efficient as possible by working with ID instead of whole object
+  const recordType = recordId.split("_")[0];
+  return recordType && ["Response", "Request", "Script"].includes(recordType);
+}
+
+function decompressRecord(record) {
+  const decompressedRecord = { ...record };
+  if (isRule(record)) {
```
wrongsahil, line 25: Missing Check for `isCompressed` key

wrongsahil (changes requested):

app/src/utils/Compression.js
nsrCodes: this is handled below on the pair level

wrongsahil (approved):

PR #1643, opened 2024-04-29
fix: implicit widget not lisiting applied rules when listener is not active

RuntimeTerror10:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/browser-extension/common/src/custom-elements/test-rule-widget/implicit-test-rule-widget/index.ts
+++ b/browser-extension/common/src/custom-elements/test-rule-widget/implicit-test-rule-widget/index.ts
@@ -36,6 +36,7 @@ class RQImplicitTestRuleWidget extends RQTestRuleWidget {
 
   addWidgetListeners() {
     this.addEventListener("new-rule-applied", (evt: CustomEvent) => {
+      this.setAttribute("applied_listener_active", "true");
       const isRuleAlreadyApplied = this.#appliedRules.some((rule) => rule.ruleId === evt.detail.appliedRuleId);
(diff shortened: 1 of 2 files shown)


browser-extension/common/src/custom-elements/test-rule-widget/implicit-test-rule-widget/index.ts
```diff
@@ -36,6 +36,7 @@ class RQImplicitTestRuleWidget extends RQTestRuleWidget {
 
   addWidgetListeners() {
     this.addEventListener("new-rule-applied", (evt: CustomEvent) => {
+      this.setAttribute("applied_listener_active", "true");
```
nafees87n, line 39: We should not use attribute for this.We should setup custom events. When the listeners are active, only then notify the widget with all the applied rules using the notify function. That way we wont have 2 sources for the applied rules

browser-extension/mv2/src/client/js/ruleExecutionHandler.js
```diff
@@ -186,9 +180,8 @@ RQ.RuleExecutionHandler.notifyRuleAppliedToExplicitWidget = (ruleId) => {
   }
 };
 
-RQ.RuleExecutionHandler.notifyRuleAppliedToImplicitWidget = (rule) => {
+RQ.RuleExecutionHandler.notifyRuleAppliedToImplicitWidget = async (rule) => {
```
nafees87n, line 183: Why mark this async?

nafees87n (approved):

PR #1664, opened 2024-05-05
[ENGG-1689] MV3: Expose onBeforeAjaxRequest hook to forward auth header in cross origin redirects

lazyvab:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.


browser-extension/mv3/src/service-worker/services/requestProcessor.ts
```diff
@@ -0,0 +1,10 @@
+interface AJAXRequestDetails {
```
nafees87n, line 1: Will also need request headers here

browser-extension/mv3/src/service-worker/services/requestProcessor.ts
```diff
@@ -0,0 +1,177 @@
...
+    const requestActionDetails = await this.processRequest(requestDetails);
+    console.timeEnd("processRequest");
+    if (!requestActionDetails) {
+      return;
+    }
+
+    switch (requestActionDetails.type) {
+      case ActionType.FORWARD_IGNORED_HEADERS:
+        console.time("forwardIgnoredHeaders");
+        await this.forwardIgnoredHeaders(tabId, requestDetails, requestActionDetails);
```
lazyvab, line 169: Going forward, there could be multiple actions to be applied in `onBeforeAJAXRequest`.
Instead of creating an action and then executing it, can we simply delegate this to a different function?
```js
public async onBeforeAJAXRequest (tabId: number, requestDetails: AJAXRequestDetails) {
  this.forwardIgnoredHeadersOnRedirect(tabId, requestDetails);
  this.handlePredefinedFunctions(tabId, requestDetails);
  // and so on...
}

private forwardIgnoredHeadersOnRedirect (tabId: number, requestDetails: AJAXRequestDetails) {
  if (!this.cachedRules.redirectRules?.length || !this.cachedRules.replaceRules?.length) {
    return;
  }

  if (!this.hasIgnoredHeadersInRequest(requestDetails.requestHeaders)) {
    return;
  }

  // check if matches any redirect/replace rule
  // apply new DNR rules
}


```

browser-extension/mv3/src/service-worker/services/requestProcessor.ts
```diff
@@ -0,0 +1,133 @@
...
+  private findMatchingRule = (rules: Rule[], url: string): Rule | undefined => {
+    return rules.find((rule) => rule.pairs.some((pair) => matchSourceUrl(pair.source, url)));
+  };
+
+  private getRedirectedUrl = (requestUrl: string, matchedRule: Rule) => {
+    const pair = matchedRule.pairs.find((pair) => matchSourceUrl(pair.source, requestUrl));
+    const redirectedUrl = pair.destination || requestUrl.replace(pair.from, pair.to);
+    if (redirectedUrl === requestUrl) {
+      return null;
+    }
```
wrongsahil, line 85: Will this work in case of wildcard and regex subsitutions?

browser-extension/mv3/src/service-worker/services/requestProcessor.ts
```diff
@@ -0,0 +1,133 @@
...
+    let removeRuleIds = [];
+
+    if (sessionRulesMap?.[requestUrl]) {
+      ruleId = sessionRulesMap[requestUrl];
+      removeRuleIds.push(ruleId);
+    }
+
+    tabService.setData(tabId, TAB_SERVICE_DATA.SESSION_RULES_MAP, {
+      ...sessionRulesMap,
+      [requestUrl]: ruleId,
```
wrongsahil, line 110: Shouldn't ruleIds be an array or map of rulesIds because there might be case when initiator_domain and auth case applies on same URL

{
    [requestUrl]: {
        auth_header: ruleId,
        initiator_domain: ruleId,
    }
}

wrongsahil (changes requested):

browser-extension/mv3/src/service-worker/services/requestProcessor.ts
nafees87n: Substitutions are not supported in case of replace rules. We need to handle only redirect rule

wrongsahil (approved):

PR #1668, opened 2024-05-07
fix: use deviceId instead of uid for uninstallation event mapping

nsrCodes:
The locally stored user id has many problems:
1. not being cleared when logged out
2. Randomly getting generated when not found (which is just generating fake new user data)
3. Mapping events from an anon user to a random uid 

shifting to using the deviceId instead


app/public/index.html
```diff
@@ -11,7 +11,8 @@
         window.location.href = 'https://app.requestly.io';
       }
 
-      if ("https://app.requestly.io/goodbye/" === window.location.href) { const a = { api_key: "62ff1b46909e50358cfca0668d41f011", events: [{ user_id: localStorage.getItem("__rq_uid") || Math.random().toString(36).slice(2, 10), event_type: "extension_uninstalled" }] }, b = new Blob([JSON.stringify(a)]); navigator.sendBeacon("https://api2.amplitude.com/2/httpapi", b) }
+      const deviceId = localStorage.getItem("__rq_device_id")
+      if ("https://app.requestly.io/goodbye/" === window.location.href && deviceId) { const a = { api_key: "62ff1b46909e50358cfca0668d41f011", events: [{ device_id: deviceId, event_type: "extension_uninstalled" }] }, b = new Blob([JSON.stringify(a)]); navigator.sendBeacon("https://api2.amplitude.com/2/httpapi", b) }
```
wrongsahil, line 15: We should use uid first and if it is not found, then we should use the deviceId.

wrongsahil, line 15: We need to set `extension_uninstall_date` attribute too on uninstall

app/public/index.html
```diff
@@ -11,7 +11,8 @@
         window.location.href = 'https://app.requestly.io';
```
wrongsahil, line 11: This file is deprecated when we shifter to vite. Please delete this file to avoid confusion in future

wrongsahil (changes requested):

app/public/index.html
nsrCodes: setting attributes is not available in the http api

wrongsahil (approved):

PR #1737, opened 2024-05-23
feat: add featureFlag to override team syncing setting

nsrCodes:
introduces two feature flags:
1. `override_team_sync_status` - default value `false`
2. `overriden_sync_status_value` - default vaule `true`

The second flag is enforced only if first one is `true`. Else the localStorage state is used


sagarsoni7 (approved):
LGTM

PR #1748, opened 2024-05-25
[ENGG-1831] fix: MV3 host and URL regex

nafees87n:
Closes issue: https://github.com/requestly/requestly/pull/1760


app/src/modules/extension/mv3RuleParser/utils.ts
```diff
@@ -60,15 +60,15 @@ const parseUrlParametersFromSource = (source: RulePairSource): ExtensionRuleCond
       case SourceOperator.MATCHES: {
         const { pattern, flags } = parseRegex(source.value);
         return {
-          regexFilter: `^https?://${pattern}(/|$)(.*)`,
+          regexFilter: `^https?://${pattern}[/?#$]?(.*)`,
```
wrongsahil, line 63: Wouldn't work if source condition is HOST+Regex `ample` and we try to open https://example.com

wrongsahil: RuleExecution shows it is executed but dnr is not applied

wrongsahil (changes requested):

app/src/modules/extension/mv3RuleParser/utils.ts
wrongsahil: How about this `^https?://[a-z0-9-]*${pattern}[a-z0-9:-]*.*`

wrongsahil (approved):

wrongsahil (approved):

wrongsahil (approved):

PR #1770, opened 2024-06-04
[ENGG-1855] fix: MV3 errors on console

nafees87n:
Closes issue: 

## 📜 Summary of changes:
1. fix: circular dependencies
2. fix: script injection errors on service worker console
3. fix: executed rules being displayed in popup when extension is disabled

## ✅ Checklist:

- [ ] Verified UI in browser.


browser-extension/mv3/src/rulesStorageService.ts
```diff
@@ -74,7 +74,7 @@ class RulesStorageService {
 
       const group = groups.find((group) => group.id === rule.groupId);
 
-      if (group.status === Status.ACTIVE) {
```
wrongsahil, line 77: This will mark the rules as disabled when the group_id doesn't exist actually. I don't know if it is the right choice as entering a non existing groupId is really common in publicApi

browser-extension/mv3/src/service-worker/services/ruleExecutionHandler.ts
```diff
@@ -3,14 +3,23 @@ import extensionIconManager from "./extensionIconManager";
...
 class RuleExecutionHandler {
-  constructor() {}
+  private isExtensionEnabled = true;
+
+  constructor() {
+    isExtensionEnabled().then((status) => (this.isExtensionEnabled = status));
+    onVariableChange<boolean>(Variable.IS_EXTENSION_ENABLED, (currentStatus) => {
+      this.isExtensionEnabled = currentStatus;
+    });
+  }
```
wrongsahil, line 23: Instead of disabling it in RuleExecutionHandler, we shouldn't even capture the request for execution in `webRequestInterceptor`

wrongsahil (changes requested):

browser-extension/mv3/src/rulesStorageService.ts
nafees87n: Should we mark the rules with non-existing groupId as Active?

PR #1879, opened 2024-06-27
[ENGG-1989] fix: DNR rules not getting unique ID while saving

nafees87n:
Closes issue: 

## ✅ Checklist:

- [ ] Verified UI in browser.

### DIFF
--- a/browser-extension/mv3/src/service-worker/services/rulesManager.ts
+++ b/browser-extension/mv3/src/service-worker/services/rulesManager.ts
@@ -33,8 +33,7 @@ const updateDynamicRules = async (options: UpdateDynamicRuleOptions): Promise<vo
       break;
     } catch (e) {
       const match = e.message.match(/Rule with id (\d+)/);
-      console.log("!!!debug", "err", e.message);
-      const ruleId = parseInt(match[1]);
+      const ruleId = match && parseInt(match[1]);
       const rqRuleId = addRules.find((rule) => rule.id === ruleId)?.rqRuleId;
 
       if (match && rqRuleId) {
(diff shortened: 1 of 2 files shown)


browser-extension/mv3/src/service-worker/services/rulesManager.ts
```diff
@@ -104,11 +107,29 @@ const addExtensionRules = async (): Promise<void> => {
...
+    } finally {
+      isApplyingRules = false;
+      processRuleApplicationQueue();
+    }
+  });
+
+  if (!isApplyingRules) {
+    processRuleApplicationQueue();
+  }
+};
```
wrongsahil, line 126: Since we remove all the DNR rules and then find the enabled rules everytime. So the end results always remains the same. So instead of making a queue, should we add debounce instead. This will make sure all the DNR aren't sent again and again to chrome (possible performance issues when setting DNRs again and again?)

wrongsahil: 1 downside i see is that when we perform bulk operation on large number of rules, rules won't  work unless all the rules `applyDynamicRules` listeners are processed and then DNRs are applied. But its the same as queue as in that case too all the rules will be removed on every `applyDynamicRules`

wrongsahil (changes requested):

browser-extension/mv3/src/service-worker/services/rulesManager.ts
nafees87n: Implemented debounce

wrongsahil (approved):
Looks good. Please do thorough testing before merging

Issue #1797, opened 2024-06-10
[ENGG-1977] MV3: 2KB limit for regexFilter in DNR


wrongsahil:
When creating host regex rules Unchecked runtime.lastError: Rule with id 1 was skipped as the "regexFilter" value exceeded the 2KB memory limit when compiled. Learn more: https://developer.chrome.com/docs/extensions/reference/api/declarativeNetRequest#regex-rules

agelbess:
Trying to save a 'replace string' rule with url>contains like '123456789012345678901234567890123456' fails.
If '12345678901234567890123456789012345' is used, then save succeeds.
It seems that it cannot handle a string longer than 35 chars

hashim-sohail:
Facing the same issue, followed the pattern 2 and it worked like a charm.

johnmanko:
I'm having this problem just matching against `yahoo.com`.

<img width="1073" height="839" alt="Image" src="https://github.com/user-attachments/assets/df723a28-fa92-4356-a9be-8863f2ff527f" />

I can't create any rules.

PR #1885, opened 2024-06-28
[ENGG-1960 ENGG-1991] feat: extension should not work on the defined blockedList

nafees87n:
Closes issue:  #1898 

## ✅ Checklist:

- [ ] Verified UI in browser.


browser-extension/mv3/src/service-worker/services/clientHandler.ts
```diff
@@ -1,12 +1,18 @@
...
+import { getBlockedDomains } from "./utils";
+import { ChangeType, onRecordChange } from "common/storage";
+import { STORAGE_KEYS } from "common/constants";
 
 const excludeMatchesPatterns = [WEB_URL, ...OTHER_WEB_URLS].map(generateUrlPattern).filter((pattern) => !!pattern);
 
+const generateBlockedDomainPattern = (hostname: string) => {
+  return `*://*${hostname}/*`;
+};
+
```
wrongsahil, line 15: - hostname -> host (as it can contain port too)
- Rename to `generateBlockedHostMatchPattern`
- This will match domains which shouldn't (`example.com` then it would match `theexample.com` too. We only need to match current domain and all subdomains)

browser-extension/mv3/src/page-scripts/ajaxRequestInterceptor/xhr.js
```diff
@@ -281,6 +282,10 @@ export const initXhrInterceptor = (debug) => {
   XMLHttpRequest.prototype.send = async function (data) {
     this.rqProxyXhr._requestData = data;
 
+    if (isRequestDomainBlocked(this.rqProxyXhr._requestURL)) {
+      send.call(this, this.rqProxyXhr._requestData);
+    }
+
     const matchedDelayRulePair = getMatchedDelayRule({
       url: this.rqProxyXhr._requestURL,
       method: this.rqProxyXhr._method,
```
wrongsahil, line 291: I don't think we need this as page_script won't be injected.

wrongsahil (changes requested):

browser-extension/mv3/src/service-worker/services/messageHandler.ts
```diff
@@ -129,6 +131,31 @@ export const initMessageHandler = () => {
...
+                message.tabUrl
+              );
+            });
+            sendResponse(isBlocked);
+          });
+        } catch (e) {
+          sendResponse(false);
+        }
+        return true;
+      }
```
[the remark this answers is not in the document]
nafees87n: https://github.com/requestly/requestly/pull/1903/files#diff-7275801d1c371aaa54b12d597703578e4d55b14ba41b2b74df397cb1ddf91b7cR134-R137

To show info in popup

browser-extension/common/src/popup/components/Popup/index.tsx
nafees87n, line 1: Complete changes are in [this PR](https://github.com/requestly/requestly/pull/1903). Ignore this file in this PR

browser-extension/mv3/src/service-worker/services/clientHandler.ts
```diff
@@ -62,23 +78,51 @@ const unregisterClientScripts = async () => {
...
+export const getBlockedDomains = async () => {
+  if (cachedBlockedDomains === null) {
+    const blockedDomains = await getRecord<string[]>(STORAGE_KEYS.BLOCKED_DOMAINS);
+    return blockedDomains ?? [];
+  }
+
+  return cachedBlockedDomains;
+};
+
```
wrongsahil, line 102: This shouldn't be here. Move to utils

browser-extension/mv3/src/service-worker/services/rulesManager.ts
```diff
@@ -89,6 +92,8 @@ const addExtensionRules = async (): Promise<void> => {
           );
         }
 
+        extensionRule.condition.excludedInitiatorDomains.push(...blockedDomains);
```
wrongsahil, line 95: Missed this in `RuleMatcher` too. Extension icon will turn green even though blocked

wrongsahil (changes requested):

browser-extension/mv3/src/service-worker/services/clientHandler.ts
```diff
@@ -74,11 +89,25 @@ export const initClientHandler = async () => {
...
+      keyFilter: STORAGE_KEYS.BLOCKED_DOMAINS,
+      changeTypes: [ChangeType.MODIFIED],
+    },
+    () => {
+      cacheBlockedDomains();
+      unregisterClientScripts().then(() => {
+        setupClientScript(isExtensionStatusEnabled);
+      });
+    }
+  );
```
wrongsahil, line 110: Can we move this to `utils.ts` and then call 
```
onBlockListChange = (callback) =>  {
    cacheBlockedDomains();
    callback();
}
```

browser-extension/mv3/src/service-worker/services/clientHandler.ts
```diff
@@ -74,11 +89,25 @@ export const initClientHandler = async () => {
   console.log("[initClientHandler]");
   const isExtensionStatusEnabled = await isExtensionEnabled();
   setupClientScript(isExtensionStatusEnabled);
+  cacheBlockedDomains();
```
wrongsahil, line 92: Instead of triggering cache ourself, let `getBlockedDomains` handle the caching if it isn't cached

browser-extension/mv3/src/utils.ts
```diff
@@ -76,3 +78,28 @@ export const debounce = (func: Function, wait: number) => {
...
+let cachedBlockedDomains: string[] = [];
+
+export const cacheBlockedDomains = async () => {
+  const blockedDomains = await getRecord<string[]>(STORAGE_KEYS.BLOCKED_DOMAINS);
+  cachedBlockedDomains = blockedDomains ?? [];
+};
+
+export const getBlockedDomains = () => {
+  return cachedBlockedDomains;
+};
```
wrongsahil, line 91:
```
export const getBlockedDomains = () => {
    If(cached) {
        return cachedBlockedDomains;
    }
    
    cacheBlockedDomains();
    return cachedBlockedDomains;
};
```

browser-extension/mv3/src/service-worker/services/tabService.ts
```diff
@@ -68,6 +70,10 @@ class TabService {
     chrome.webNavigation.onCommitted.addListener((navigatedTabData) => {
       if (navigatedTabData.frameId === 0) {
         this.resetPageData(navigatedTabData.tabId);
+        const isBlocked = isUrlInBlockList(navigatedTabData.url);
+        if (isBlocked) {
+          extensionIconManager.markExtensionBlocked(navigatedTabData.tabId);
+        }
```
wrongsahil, line 76: This should be in extensionIconManager `chrome.tabs.onUpdated` as this introduces cirular loop

wrongsahil (changes requested):

browser-extension/mv3/src/service-worker/services/clientHandler.ts
nafees87n: `cacheBlockedDomains()` is asynchrounous so if `getBlockedDomains()` is made to handle caching then `getBlockedDomains` has to return a promise. 
It will be better if `getBlockedDomains()` does not return promise because of its usage in `webRequestInterceptor`.

nafees87n: If we trigger the caching ourself then there would be no instance when blocked domains have not been cached

browser-extension/mv3/src/utils.ts
wrongsahil: This is still missing. How would the cache be updated first time?

browser-extension/mv3/src/service-worker/services/clientHandler.ts
wrongsahil: There can be still be some instance i think if webrequestinterceptor is initialized before this `cacheBlockedDomains` is still in the promise queue. So in this case too, the cachedBlockedDomains isn't ready and webRequestInterceptor bypasses it
But this looks weird and out the place as this method responsibility shouldn't be to initialize the cache. It should be in the concerned file only.

If we initialize it in getBlockedDomains only, there is very small chance that the blockedDomains aren't cached (and that too when extension service worker is init by a website that is in the blocklist).

wrongsahil (changes requested):

wrongsahil (approved):

wrongsahil:
This might cause issues with https://github.com/requestly/requestly/issues/1797 if there are lots of domains added as the 2kb limit is on DNR rule not regex condition only

PR #1918, opened 2024-07-05
[ENGG-1969]: incentivization enhancements

rohanmathur91:
Context:
- Combined create and test rule together (only for users who join after this release, for existing users no change in events), user will only get the credits when they test the rule.
- Few analytics fixes


app/src/components/features/rules/TestThisRule/components/TestRuleHeader/index.tsx
```diff
@@ -37,6 +38,7 @@ export const TestRuleHeader = () => {
   const { sheetPlacement } = useBottomSheetContext();
 
   const { claimIncentiveRewards } = useIncentiveActions();
+  const isNewUserForIncentivization = useIsNewUserForIncentivization("2024-07-05");
```
RuntimeTerror10, line 41: The release date string literal is being used in every instance of `useIsNewUserForIncentivization`. Can we create a constant for release dates for incentivization and use that instead.

RuntimeTerror10: If in future any more enhancements are made for new users then we can just add a update or add a new date in the constant.

RuntimeTerror10 (approved):

app/src/components/features/rules/TestThisRule/components/TestRuleHeader/index.tsx
rohanmathur91: Okay

PR #1956, opened 2024-07-17
[ENGG-2050]: add rules details panel in editor

rohanmathur91:
https://github.com/user-attachments/assets/f336aaaf-0a98-45ca-b9a8-f4862ec717bf


app/src/store/initial-state/index.js
```diff
@@ -34,6 +34,7 @@ const INITIAL_STATE = {
       data: false,
       config: false,
       hasUnsavedChanges: false,
+      showDetailsPanel: true,
```
RuntimeTerror10, line 37: Why are we keeping the details panel visibility in redux store instead of maintaining the state at local level?

rohanmathur91: Its one time thing, after close we should not show it again on new rule creation

app/src/views/features/rules/RuleEditor/components/RuleDetailsPanel/RuleDetailsPanel.tsx
```diff
@@ -0,0 +1,119 @@
...
+                fill="#8F8F8F"
+              />
+            </svg>
+            <svg xmlns="http://www.w3.org/2000/svg" width="68" height="5" viewBox="0 0 68 5" fill="none">
+              <ellipse opacity="0.1" cx="34.0001" cy="2.52071" rx="33.0274" ry="1.98165" fill="#8F8F8F" />
+            </svg>
```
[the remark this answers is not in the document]
rohanmathur91: It is specific to this component only, so instead of keeping in top level assets keeping it here only

RuntimeTerror10 (approved):

PR #1988, opened 2024-07-23
[ENGG-2053]: update applied rules implicit widget

rohanmathur91:


browser-extension/common/src/custom-elements/test-rule-widget/index.ts
```diff
@@ -62,13 +80,14 @@ export abstract class RQTestRuleWidget extends RQDraggableWidget {
...
-            <div id="logo-container"> 
+            <div id="logo-container">
                 <span id="heading-logo">${RQLogo}</span>
                 <span id="logo-text">requestly</span>
             </div>
             <div id="actions-container">
-                <button id="settings-button" class="hidden">${SettingsIcon}</button>
+                <button id="settings-button" class="hidden" data-tooltip="Hide widget in app settings">${VisibilityIcon}</button>
                 <button id="minimize-button">${MinimizeIcon}</buttton>
+                <button id="close-button" class="hidden">${CloseIcon}</buttton>
```
nafees87n, line 90: This is common to both the explicit and implicit widgets. Can we keep the components being used in only one of them specfic to its class? It will be cleaner instead of managing its visibility using javascript

rohanmathur91: Since its a close button which we might be used in both type of widgets, so that’s why keeping it here only.

nafees87n (approved):

PR #2026, opened 2024-07-29
[ENGG-2113] fix: status toggle does not save the updated rule definition

nafees87n:
## 📜 Summary of changes:

1. Moved rule sanitization and script rule validations inside `saveRule` function
2. Common function is called when rule is toggled and save button is clicked


app/src/views/features/rules/RuleEditor/components/Header/ActionButtons/actions/index.js
```diff
@@ -9,12 +9,29 @@ import Logger from "lib/logger";
...
+  //Syntactic Validation
+  const syntaxValidatedRule = await validateSyntaxInRule(dispatch, fixedRuleData);
+
+  if (!syntaxValidatedRule) {
+    return;
+  }
+
+  const parsedRuleData = syntaxValidatedRule || fixedRuleData;
+
+  ruleToSave = migrateRuleToMV3(parsedRuleData).rule;
```
nsrCodes, line 34: moving these checks here is also making them run on template import and status toggle. We already have an unknown performance issue in status toggle which makes it take unexpectedly long times, I would not want to add unnecessary operations to worsen the experience

nsrCodes: is this absolutely necessary?

app/src/views/features/rules/RuleEditor/components/Header/ActionButtons/Status/index.js
```diff
@@ -39,19 +42,16 @@ const Status = ({ isDisabled = false, location, isRuleEditorModal }) => {
     const isCreateMode = location.pathname.indexOf("create") !== -1;
 
     !isCreateMode &&
-      StorageService(appMode)
-        .saveRuleOrGroup(
-          {
-            ...currentlySelectedRuleData,
-            status: newValue,
-          },
-          false
```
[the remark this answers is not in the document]
nsrCodes: - `lastModified` should not change on just status toggle. 
- if we save new changes except status toggle, then yes `lastModified` should update.
- but here we are creating inconsistent behaviour:
  - status toggle from the list will not save rule => will not update `lastModified`
  - status toggle from editor (based on these changes) will save the rule => will updated `lastModified`
- Ideally, status toggle should only update the status attribute, and [should not  be related to `unsavedChanges`](https://github.com/requestly/requestly/pull/2026/files#diff-b8dd3088b2642ca9bf2573283273cf520f62a3219437f642c3c2c15d10a23817R53-R54) state
  - if there are any unsaved changes in the editor the save button has an astreik _(*)_ and the user will also be [prompted by the browser](https://github.com/requestly/requestly/blob/fd50de3d75639c2872fb8508e1b8de3ff5a8668f/app/src/views/features/rules/RuleEditor/RuleEditor.js#L53-L56). 
  - Which means ideally the code should reflect this behaviour => code for validation of unsaved changes happens on save rule button only

nsrCodes (approved):
Discussed that save on toggle has been a conscious default behaviour, hence previous comments are obsolete. Just add a comment in the code making it obvious. LGTM!

PR #2064, opened 2024-08-07
ENGG-2072: Automate add users to email campaigns

RuntimeTerror10:


app/src/features/pricing/components/PricingTable/components/PlanColumn/index.tsx
```diff
@@ -96,11 +95,21 @@ export const PlanColumn: React.FC<PlanColumnProps> = ({
...
 
+  const handlePricingQuantityChanged = useCallback(() => {
+    const addToApolloSequence = httpsCallable(getFunctions(), "pricing-addToApolloPricingFiddleSequence");
+    addToApolloSequence().catch((error) => {
+      Logger.log("Error adding user to apollo sequence", error);
+    });
+  }, []);
+
+  const debouncedHandlePricingQuantityChanged = useDebounce(handlePricingQuantityChanged, 10000);
+
```
nsrCodes, line 106: debouncing like this holds the risk of loosing the task when tab closes. Can we do something better? We should at least reduce the debounce time since 10s is a lot!

nsrCodes (approved):

PR #2072, opened 2024-08-09
[ENGG-2136] feat: sharedState in extension dynamic request, response rules

nafees87n:
Implements #2094 

Dynamic JS type in request/response can make use of `$sharedState` variable to access data across different requests and response. Currently, it is implemented in the tab's context i.e the `$sharedState` will be persisted across a tab session.

## Storage
- `sharedState` is stored is window's RQ namespace. 
- Injected using clientHandler on pageNavigation
- On page navigation, `sharedState` is cached in `tabService` to be retrieved after navigation

## Usage with iframes
- Cross origin iframes cannot access window.top, in that case the iframe uses a different `sharedState`
https://github.com/requestly/requestly/blob/fd19b6ec59271148f0f641b9abe25bdb38458c2b/browser-extension/mv3/src/page-scripts/ajaxRequestInterceptor/fetch.js#L183-L187

## Accessing `sharedState` in dynamic js function in request/response rule
`sharedState` is passed by reference to the function just like other request/response args 
https://github.com/requestly/requestly/blob/fd19b6ec59271148f0f641b9abe25bdb38458c2b/browser-extension/mv3/src/page-scripts/ajaxRequestInterceptor/utils.js#L7-L10

## Usage
image


browser-extension/mv3/src/page-scripts/ajaxRequestInterceptor/utils.js
```diff
@@ -1,12 +1,12 @@
...
+import { EXTENSION_MESSAGES, PUBLIC_NAMESPACE } from "common/constants";
 import { matchRuleWithRequest, matchSourceUrl } from "../../common/ruleMatcher";
 import { SourceKey, SourceOperator } from "common/types";
 
 let logShown = false;
 
 export const getFunctionFromCode = (code, ruleType) => {
   try {
-    return new Function("args", `return (${code})(args);`);
+    return new Function("args,sharedState", `return (${code})(args,sharedState);`);
```
nsrCodes, line 9: @wrongsahil @nafees87n I think now would be a good time to implement the same approach in both proxy and extension for how we handle function code strings.

because in the proxy we do it slightly differently _(following is before the [sharedState changes](https://github.com/requestly/requestly-proxy/blob/12354850a84817a73ea27d9c0ede472e8f10cc1b/src/utils/index.ts#L5-L7))_:

```typescript
export const getFunctionFromString = function (functionStringEscaped) {
    return new Function("return " + functionStringEscaped)();
};
```

This generates the function that is then called as `userFunction(args)`

> The difference is where the scope of `args` is. Its trivial but good to keep it consistent across both the implementations

wrongsahil, line 9: Make it `$sharedState` instead to keep consistent with desktop app

browser-extension/mv3/src/page-scripts/ajaxRequestInterceptor/fetch.js
```diff
@@ -179,7 +180,13 @@ export const initFetchInterceptor = (debug) => {
           };
         }
 
-        customResponse = getFunctionFromCode(responseModification.value, "response")(evaluatorArgs);
+        try {
+          sharedState = window.top[PUBLIC_NAMESPACE]?.sharedState ?? {};
+        } catch (e) {
+          sharedState = window[PUBLIC_NAMESPACE]?.sharedState ?? {};
+        }
```
wrongsahil, line 187: We can fetch this directly in `getFunctionFromCode` from window....sharedState

browser-extension/mv3/src/service-worker/services/globalStateManager.ts
```diff
@@ -0,0 +1,52 @@
...
+      world: "MAIN",
+    });
+  }
+
+  constructor() {
+    // this.variables = {};
+  }
+
+  updateSharedStateInStorage(tabId: number, sharedState: Record<string, any>) {
+    tabService.setData(tabId, "sharedState", sharedState);
```
wrongsahil, line 31: Use constant for "sharedState" -> TAB_SERVICE_DATA.SHARED_STATE

wrongsahil (changes requested):

browser-extension/mv3/src/page-scripts/ajaxRequestInterceptor/utils.js
nsrCodes: @nafees87n please refer the changes in [this file](https://github.com/requestly/requestly-proxy/blob/master/src/utils/index.ts) for the desktop implementation

browser-extension/mv3/src/page-scripts/ajaxRequestInterceptor/fetch.js
```diff
@@ -179,7 +181,11 @@ export const initFetchInterceptor = (debug) => {
           };
         }
 
-        customResponse = getFunctionFromCode(responseModification.value, "response")(evaluatorArgs);
+        const isUserFunctionValid = getFunctionFromCode(responseModification.value, "response");
+
+        if (isUserFunctionValid) {
+          customResponse = generateUserFunctionWithSharedState(responseModification.value, debug)(evaluatorArgs);
+        }
```
wrongsahil, line 188: Why can't we just use `getFunctionFromCode` (which implicitly contains shared state in the closure). Why do we need to check whether function is valid first?

wrongsahil (changes requested):

browser-extension/mv3/src/page-scripts/ajaxRequestInterceptor/fetch.js
nafees87n: Kept it separate that so that `notifyOnErrorOccurred` is not called each time there's `generateUserFunctionWithSharedState`. There could be instances when we do not want to call `notifyOnErrorOccurred` intentionally

wrongsahil (approved):

PR #2083, opened 2024-08-16
ENGG-2122: Add share action for rule group in rules table

RuntimeTerror10:


app/src/features/rules/utils/index.ts
nsrCodes, line 1: These changes don't seem required. `StorageRecord` has `objectType` for both RULE and GROUP

app/src/features/rules/screens/rulesList/components/RulesList/components/RulesTable/utils/rules.ts
```diff
@@ -96,3 +99,22 @@ export const checkIsRuleGroupDisabled = (allRecordsMap: Record<string, StorageRe
...
+          (record: StorageRecord) => isRule(record) && record?.groupId === groupId
+        );
+        resolve(groupRules as Rule[]);
+      })
+      .catch((error) => {
+        Logger.log(error);
+        resolve([]);
+      });
+  });
+};
```
nsrCodes, line 120: I don't think there is one specific file for such utils, but I believe it would make more sense to move it to [`src/utils/rules/misc.js`](https://github.com/requestly/requestly/blob/48cd10fb7eb2feb95787c217be1f7d68f59711b8/app/src/utils/rules/misc.js)

nsrCodes (approved):

app/src/features/rules/utils/index.ts
nsrCodes: tested, leads to bugs if not added

Issue #2058, opened 2024-08-06
Support to append multiple header values for a single HTTP header using header rule


nafees87n:
Requestly supports adding multiple header values for single header name using the `Add` operation but due to some bug it always overrides the existing header value.

Multiple header values separated by delimiter are supported in HTTP headers like 

```
Cookie: name=value; name2=value2; name3=value3
Accept: text/html, application/xhtml+xml, application/xml;q=0.9, image/webp, */*;q=0.8
```

# Current behaviour

Currently, header rule supports 3 operations:

1. `Add Operation` - Creates a new header with the specified name and value. If a header with the same name already exists, this operation adds another header with the same name rather than replacing the existing one. This allows for multiple headers with the same name but different values.
**But due to a bug, the existing header is always replaced with the new value.**

2. `Override Operation` - Always overrides the value of the specified header if present in the request/response.
**But due to a bug, the new header value is always set even if the specified header is not present in the request/response**

3. `Remove Operation` - Removes the specified header from the request/response if present.

# Proposed behaviour

The header rule can modelled as per the [DNR operations](https://developer.chrome.com/docs/extensions/reference/api/declarativeNetRequest#type-HeaderOperation) supported in the chrome API.
DNR header action supports 3 operations:

1. `append` - Adds a new entry for the specified header. This operation only supports standard HTTP request headers that can specify multiple values for a single entry are supported otherwise API throws error while saving. Although no such restriction is present for response headers.

2. `set` - Sets a new value for the specified header, removing any existing headers with the same name.

3. `remove` - Removes all entries for the specified header.

nafees87n:
# List of supported standard HTTP request headers that can specify multiple values for a single entry

1. accept
2. accept-encoding
3. accept-language
4. access-control-request-headers
5. cache-control
6. connection
7. content-language
8. cookie
9. forwarded
10. if-match
11. if-none-match
12. keep-alive
13. range
14. te
15. trailer
16. transfer-encoding
17. upgrade
18. via
19. want-digest
20. x-forwarded-for

Please refer [here](https://chromium-review.googlesource.com/c/chromium/src/+/3922770/11/extensions/browser/api/declarative_net_request/constants.h#258)

nafees87n:
I can see a bug in the chromium implementation.

[Header names are case-insensitive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers) but DNR APIs match the case as well.

### Example

[...]

PR #2085, opened 2024-08-16
[ENGG-2160] feat: added append support in DNR parser for header rule

nafees87n:
Implements #2058 

- `Add` operation appends the header whenever possible otherwise works like `override` (sets the header). Possible for response headers and[ list of possible request headers](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/declarativeNetRequest/ModifyHeaderInfo#header_limits).
- `Override` operation overrides the header value and sets the value if the header does not exist.


app/src/modules/extension/mv3RuleParser/parseHeadersRule.ts
```diff
@@ -14,6 +40,16 @@ const parseHeaders = (headers: HeadersRuleModificationData[]): ModifyHeaderInfo[
...
+      } else if (
+        header.type === HeaderRuleActionType.ADD &&
+        (headerType === "Response" ||
+          (headerType === "Request" && APPEND_SUPPORTED_HEADERS.includes(header.header.toLowerCase())))
+      ) {
+        return {
+          header: header.header.toLowerCase(),
+          value: header.value,
+          operation: "append" as HeadersRuleOperation,
+        };
```
wrongsahil, line 52: Can we add a comment here on why are we doing this

wrongsahil (approved):

PR #2134, opened 2024-09-02
fix: remove unwanted 'allRules' from useEffect dependency array

nsrCodes:
removed `allRules` from `useEffect` dependency array. it was causing infinite rerenders in case of insert script rule.


nafees87n (approved):
LGTM

PR #2156, opened 2024-09-09
[ENGG-2176] fix: drag and drop - row drop UX

rohanmathur91:
https://github.com/user-attachments/assets/6508ac51-a58c-4578-956a-f4fa14d735e4


rohanmathur91:
Not going with this approach for now.

PR #2164, opened 2024-09-12
[ENGG-2280] feat: display bluecore events

rohanmathur91:


browser-extension/common/package.json
```diff
@@ -56,9 +56,11 @@
     "@devtools-ds/object-inspector": "^1.2.1",
     "@devtools-ds/table": "^1.2.1",
     "@devtools-ds/themes": "^1.2.1",
+    "@types/lodash": "^4.17.7",
```
nafees87n, line 59: Move types to devDependencies

browser-extension/common/src/devtools/containers/analytics-inspector/components/VendorEvent/VendorEvent.tsx
nafees87n, line 1: Can we name it something like `VenderEventComponent` or something else? Currently it does not look like a component

rohanmathur91: Its already inside `/components/` so naming it `VenderEventComponent` will go against convention, so keeping it as it is for now.

nafees87n: What about `VendorEventDisplay` or something that the component represents? `VendorEvent` is a noun specfic to events, we also have state variable names as `vendorEvents`, There will be types related to `vendorEvent`. Having a component named such will create confusion.

nafees87n (approved):

PR #2215, opened 2024-09-27
[ENGG-2312]: move api-client into feature folder

rohanmathur91:


app/src/features/apiClient/screens/apiClient/components/client-view/APIClientView.tsx
RuntimeTerror10, line 1: Rename this file's parent folder to `clientView`

app/src/components/misc/Favicon/index.tsx
```diff
@@ -1,7 +1,7 @@
 import { Avatar } from "antd";
 import React, { CSSProperties, useEffect, useMemo, useState } from "react";
-import { addUrlSchemeIfMissing } from "views/features/api-client/apiUtils";
 import { debounce } from "lodash";
+import { addUrlSchemeIfMissing } from "features/apiClient/screens/apiClient/apiUtils";
```
RuntimeTerror10, line 4: This is being used outside the APIClient's feature folder. Can we export it from the feature folder's root index.ts file?

app/src/components/common/APIClient/APIClient.tsx
RuntimeTerror10, line 1: This is one of the APIClient modals, which should be inside the feature folder  in `modals` folder.

RuntimeTerror10 (approved):

app/src/components/misc/Favicon/index.tsx
rohanmathur91: Ideally this should not be here, for now leaving this as it is, adding a todo.

app/src/components/common/APIClient/APIClient.tsx
rohanmathur91: Moved into `apliClient/components/common`, not adding into modals, since it can also be used as a screen

PR #2219, opened 2024-10-03
[ENGG-2316] added environment variable helper methods

nafees87n:


app/src/backend/environmentVariables/hooks/useEnvironmentVariables.ts
```diff
@@ -0,0 +1,98 @@
...
+    const variableQuery = doc(db, "environmentVariables", ownerId);
+
+    return onSnapshot(variableQuery, (doc) => {
+      if (doc.exists()) {
+        const variables: Record<string, string> = doc.data()[environment];
+
+        if (variables) {
+          const newVariables = Object.entries(variables).reduce((acc, [key, value]) => {
+            acc[key] = {
+              syncValue: value,
```
wrongsahil, line 38: We should add `type=string|number` too both here as well in firestore

app/src/backend/environmentVariables/hooks/useEnvironmentVariables.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const setVariable = async (key: string, value: EnvironmentVariableValue) => {
+    const newVariable: Record<string, EnvironmentVariableValue> = {
+      [key]: {
+        localValue: value.localValue,
+        syncValue: value.syncValue,
+      },
+    };
+
+    dispatch(environmentVariablesActions.setVariables({ newVariables: newVariable, environment }));
```
wrongsahil, line 65: Dispatch should happen after the values are set in firestore successfuly. Else it should revert.

app/src/backend/environmentVariables/hooks/useEnvironmentVariables.ts
```diff
@@ -0,0 +1,98 @@
...
+        if (variables) {
+          const newVariables = Object.entries(variables).reduce((acc, [key, value]) => {
+            acc[key] = {
+              syncValue: value,
+            };
+            return acc;
+          }, {} as Record<string, EnvironmentVariableValue>);
+
+          dispatch(environmentVariablesActions.setVariables({ newVariables, environment }));
+        }
```
wrongsahil, line 44: All the firebase/networking related code can be moved to a single file `index.js` here. So this onSnapshot listener can be moved there too

app/src/store/features/environmentVariables/slice.ts
```diff
@@ -0,0 +1,58 @@
...
+                {
+                  localValue: value.localValue ?? prevValue?.localValue,
+                  syncValue: value.syncValue ?? prevValue?.syncValue,
+                },
+              ];
+            })
+          ),
+        },
+      };
+    },
```
wrongsahil, line 40: Lets create the reducer file separately and user immer so that we don't need to spread the envValues

app/src/store/features/environmentVariables/slice.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const slice = createSlice({
+  name: ReducerKeys.ENVIRONMENT_VARIABLES,
+  initialState,
+  reducers: {
+    resetState: () => initialState,
+    setVariables: (
+      state,
+      action: PayloadAction<{
+        newVariables: Record<string, EnvironmentVariableValue>;
```
wrongsahil, line 18: Make `Record<string, EnvironmentVariableValue>` as a type `EnvVar` so that we can use it everywhere for consistency

app/src/backend/environmentVariables/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  return doc(db, "environmentVariables", ownerId);
+};
+
+export const setEnvironmentVariable = async (
+  ownerId: string,
+  payload: {
+    newVariable: {
+      key: string;
+      value: string | number | boolean;
+    };
```
wrongsahil, line 15: Use the same type `Record<string, EnvironmentVariableValue>` and make `localValue` optional so that it isn't synced to firebase

wrongsahil (changes requested):

app/src/backend/environmentVariables/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    getDocPath(ownerId),
+    {
+      [payload.environment]: {
+        ...newVariables,
+      },
+    },
+    { merge: true }
+  );
+};
+
```
wrongsahil, line 31: Each environment should be a separate doc in the collection `envs`

`environmentVariables/{ownerId}/envs/{envName}`

```
name: EnvName,
variables: {

}
```

wrongsahil (changes requested):

app/src/store/features/environment/types.ts
```diff
@@ -0,0 +1,10 @@
+import { EnvironmentVariable } from "backend/environment/types";
+
+export interface InitialState {
+  currentEnvironment: EnvironmentDetails;
+  variables: EnvironmentVariable;
+}
```
wrongsahil, line 6: We should keep a map of all environment and their variables. Since user will be switching between them, the local values will get cleared and will need to be set everytime

wrongsahil:
```
export interface InitialState {
  currentEnvironment: EnvironmentDetails;
  envs: {
      [envName]: {
          variables: {
....
          }
      }
  }
}
```

wrongsahil (changes requested):

app/src/backend/environment/hooks/useEnvironmentManager.ts
```diff
@@ -0,0 +1,129 @@
...
+        dispatch(environmentVariablesActions.removeVariableFromEnvironment({ key, environmentId }));
+      })
+      .catch((err) => {
+        toast.error("Error while removing environment variables.");
+        Logger.error("Error while removing environment variables from db", err);
+        console.error("Error while removing environment variables from db", err);
+      });
+  };
+
+  const renderString = <T>(template: string | Record<string, any>): T => {
```
wrongsahil, line 109: Rename this to `renderWithVars()` or `render` only as this is rendering not only string but object types too

wrongsahil (changes requested):

app/src/backend/environment/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export const attatchEnvironmentVariableListener = (
+  ownerId: string,
+  callback: (newVariables: EnvironmentMap) => void
+) => {
+  if (!ownerId) {
+    return () => {};
+  }
+
+  const variableDoc = collection(db, "environmentConfigs", ownerId, "environments");
```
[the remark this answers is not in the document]
nafees87n: We will need to fetch all the environments once and then attach listener to the current env

app/src/features/apiClient/screens/apiClient/APIClient.tsx
```diff
@@ -63,10 +65,17 @@ export const APIClient: React.FC<Props> = () => {
...
+    (apiEntry: RQAPI.Entry) => {
+      const renderedEntry = {
+        ...apiEntry,
+        request: renderVariables<RQAPI.Entry["request"]>(apiEntry.request),
+      };
+      setHistory((history) => [...history, renderedEntry]);
+      addToHistoryInStore(renderedEntry);
+    },
+    [renderVariables]
+  );
```
wrongsahil, line 78: Add to history shouldn't render template again. It should get the rendered API Request directly

wrongsahil (approved):

PR #2220, opened 2024-10-03
[ENGG-2316] feat: integrate environment variables with API client

nafees87n:
- PR https://github.com/requestly/requestly/pull/2219 adds support for env variables.
- This PR integrates with API client 
- `handlebars` is used to render template


app/src/backend/environmentVariables/hooks/useEnvironmentVariables.ts
```diff
@@ -85,13 +78,26 @@ const useEnvironmentVariables = () => {
     return removeEnvironmentVariable(ownerId, { environment, key });
   };
 
+  const renderVariables = (template: string) => {
```
wrongsahil, line 81: Lets rename this to `renderString` and create another method `render` which can render JSON object while recursively going through every keys.

app/src/backend/environmentVariables/utils.ts
```diff
@@ -0,0 +1,18 @@
...
+    acc[key] = value.localValue ?? value.syncValue;
+    return acc;
+  }, {} as Record<string, string | number | boolean>);
+  try {
+    return lodash.template(template)(parsedVariable);
+  } catch (e) {
+    Logger.error("Error while rendering template");
+    return template;
+  }
+};
```
wrongsahil, line 18: We should use backbone here too as it is more extensible and we're already using it in mock server

https://github.com/requestly/requestly-mock-server/blob/1831d97d65eb335548090f0a0faa5b28936f6f90/src/core/utils/templating/index.ts#L8-L15

app/src/features/apiClient/screens/apiClient/components/clientView/APIClientView.tsx
```diff
@@ -71,6 +76,54 @@ const APIClientView: React.FC<Props> = ({ apiEntry, notifyApiRequestFinished })
...
+        return {
+          id: headerKeyValue.id,
+          key: renderVariables(headerKeyValue.key),
+          value: renderVariables(headerKeyValue.value),
+        };
+      });
+      parsedRequest.headers = parsedHeaders;
+
+      if (request.body) {
+        const parsedBody =
```
wrongsahil, line 108: Instead of all this logic, can we create a common render method in EnvVariableService which recursively renders a JSON object?

app/src/features/apiClient/screens/apiClient/components/clientView/APIClientView.tsx
```diff
@@ -208,14 +261,16 @@ const APIClientView: React.FC<Props> = ({ apiEntry, notifyApiRequestFinished })
       sanitizedEntry.request.body = removeEmptyKeys(sanitizedEntry.request.body as KeyValuePair[]);
     }
 
+    const parsedRequest = parseRequestData(sanitizedEntry.request);
```
wrongsahil, line 264: Rename this to `renderedRequest`

PR #2225, opened 2024-10-04
chore: show alert when trying to save too many network logs

nsrCodes:
since we can't know for sure because 
1. it is more about the size of the logs, rather than the number of entries
2. the threshold might be different for machines with different specifications

starting with a tested estimate of 1500 logs. will change later if required

Screenshot 2024-10-04 at 9 41 40 AM

### DIFF
--- a/app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/ActionHeader.jsx
+++ b/app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/ActionHeader.jsx
@@ -176,14 +176,15 @@ const ActionHeader = ({
   const showFilterLogsAlert = (logsToSave, src) => {
     trackSavingTooManyLogsAlertShown(logsToSave.log.entries.length ?? undefined, src);
     Modal.error({
-      title: "Too many logs",
+      title: "Log Limit Exceeded",
       content: (
         <Typography>
-          Trying to save these many logs might lead to unexpected behavior.
+          Saving this many logs may lead to unexpected app behaviour.
           <br />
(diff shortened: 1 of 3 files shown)


nsrCodes:
updated copy and width of modal

Screenshot 2024-10-04 at 10 02 28 AM

app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/ActionHeader.jsx
```diff
@@ -167,6 +173,21 @@ const ActionHeader = ({
     trackRQDesktopLastActivity(TRAFFIC_TABLE.TRAFFIC_TABLE_FILTER_CLICKED);
   };
 
+  const showFilterLogsAlert = (logsToSave, src) => {
```
RuntimeTerror10, line 176: Can we rename this to `showExcessLogsAlert`?

app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/ActionHeader.jsx
```diff
@@ -167,6 +173,21 @@ const ActionHeader = ({
...
+    trackSavingTooManyLogsAlertShown(logsToSave.log.entries.length ?? undefined, src);
```
RuntimeTerror10, line 177: props like `logsCount` and `filteredLogsCount` are passed to this component, can  we use theses instead of using `.length`. 

Same for `canSaveLogsWithoutCrashing` function

RuntimeTerror10 (approved):

PR #2229, opened 2024-10-08
ENGG-2335: Prefill operationName filter in Modify Response for GraphQL

RuntimeTerror10:


app/src/components/common/RuleEditorModal/index.tsx
```diff
@@ -108,13 +107,38 @@ const RuleEditorModal: React.FC<props> = ({ isOpen, handleModalClose, analyticEv
         status: Status.ACTIVE,
       };
 
+      if (ruleType === GLOBAL_CONSTANTS.RULE_TYPES.RESPONSE) {
+        const requestBody = ruleData.request.body ? JSON.parse(ruleData.request.body) : undefined;
+        // Handling prefill for graphql resource type response rule
+        if (ruleData.request.path.indexOf("/graphql") !== -1 && requestBody[0] && requestBody[0].operationName) {
```
nafees87n, line 113: Lets not check for `/graphql`. The path could vary, some sites also use `/gql`

nafees87n, line 113: Request will not be an array always.
Network logs already have `graphQLDetails`[ metadata](https://github.com/requestly/requestly/blob/e25d8a9f13a6730813eab9fd06c72e2f278c627b/browser-extension/common/src/devtools/containers/network/networkLogUtils.ts). We should use that in the editorModal as well.

`if (graphQLDetails.operationName!==null)`
```
              "requestPayload": {
                            "key": "operationName",
                            "operator": "Equals",
                            "value": <graphQLDetails.operationName>
                        }
```

`else if(graphQLDetails.query!==null)`
```
"requestPayload": {
                            "key": "query",
                            "operator": "Contains",
                            "value": "graphQLDetails.query"
                        }
```

This won't cater to payloads of type Array but thats not a popular usecase. We can leave it for now.

nafees87n (changes requested):

nafees87n (approved):

PR #2232, opened 2024-10-10
Chore: Add feature compatibility check for $1 support in response rule

RuntimeTerror10:


nsrCodes (approved):
Change version to 1.7.6

PR #2233, opened 2024-10-10
feat: bundle requestly-core as both cjs and esm

nsrCodes:
- had to do this to make it easily importable inside requestly-proxy (esm modules cannot run inside electron renderer)
- supporting PRs to work with this change are:
    - https://github.com/requestly/requestly-proxy/pull/52
    - https://github.com/requestly/requestly-desktop-app/pull/112

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -3,7 +3,7 @@
  * Right now the imports are defined using require, once changed to ES6 module we can move to import module
  */
 if (typeof isReactApp === "undefined") {
-  var isReactApp = document.getElementById("root") !== null;
+  var isReactApp = document?.getElementById("root") !== null;
 }
 const CONSTANTS = {};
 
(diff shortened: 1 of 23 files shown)


common/constants.js
```diff
@@ -228,9 +228,8 @@ CONSTANTS.REQUEST_STATE = {
   COMPLETE: "COMPLETE",
 };
 
-if (isReactApp) {
-  module.exports = CONSTANTS;
-} else {
+export default CONSTANTS;
+if (!isReactApp) {
   /** For legacy apps- browser extension */
   Object.assign(window.RQ, CONSTANTS);
```
nafees87n, line 234: Note: This is only consumed by MV2.

nafees87n (approved):
LGTM

PR #2239, opened 2024-10-15
ENGG-2326: Variables listing Screen

RuntimeTerror10:
Screenshot 2024-10-15 at 3 47 33 PM


app/src/backend/environment/hooks/useEnvironmentManager.ts
```diff
@@ -18,8 +18,9 @@ import { toast } from "utils/Toast";
 
 let unsubscribeListener: () => void = null;
 
-const useEnvironmentManager = () => {
+const useEnvironmentManager = (initListenerAndFetcher: boolean = false) => {
```
nafees87n, line 21: Instead of `initListenerAndFetcher` what if we check if listener has been attached or not and attach it only once?

app/src/backend/environment/types.ts
```diff
@@ -13,3 +13,9 @@ export interface EnvironmentData {
 }
 
 export type EnvironmentMap = Record<string, EnvironmentData>;
+
+export enum VariableType {
```
nafees87n, line 17:
```suggestion
export enum EnvironmentVariableType {
```

app/src/features/apiClient/screens/apiClient/components/variablesListView/components/VariablesList/VariablesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+          if (variable.key) {
+            acc[variable.key] = {
+              type: variable.type,
+              syncValue: variable.syncValue,
+              localValue: variable.localValue,
+            };
+          }
+          return acc;
+        }, {});
+
```
nafees87n, line 53: Do not saving of two variables with same name. Show an error in UI

PR #2257, opened 2024-10-24
[ENGG-2374]: Design System - Breadcrumbs with Record Name update support

rohanmathur91:
https://github.com/user-attachments/assets/cf2689f7-50a2-4804-80af-835b7d516c73


app/src/features/rules/routes.tsx
```diff
@@ -50,35 +50,79 @@ export const ruleRoutes: RouteObject[] = [
   {
     path: `${PATHS.RULES.INDEX}`,
     element: <RulesContainerV2 />,
+    handle: {
+      breadcrumb: {
+        label: "Rules",
+        navigateTo: PATHS.RULES.INDEX,
```
wrongsahil, line 56: Instead of navigateTo, can't we just use the path?

wrongsahil (approved):
This looks awesome. Great work @rohanmathur91

app/src/features/rules/routes.tsx
rohanmathur91: Yes, updated

PR #2287, opened 2024-11-08
Fix cd to common cmd

Permanently:
Because for some reason, it was trying to back out of two directories instead of one.


Permanently:
https://github.com/requestly/requestly/commit/28032cdfd5e7f886b22ed45c8c1613db8e0f85cc#diff-3bc411121681ddc6909bc8b987e8b06ba97f29ba117dca1451433a8feaeb9510R4 finally addressed this. Shame I never got the credit for pointing it out first. :|

Issue #2288, opened 2024-11-10
bug: unable to install using install script


nkg447:
### Describe your issue?

install script defined does not have proper order of modules in which it should be installed. This caused issues for people who want to get started quickly. 

Error while running `install.sh`

```
***** Installing analytics vendor's dependencies *****
install.sh: line 6: cd: ../../common/analytics-vendors: No such file or directory
```
```
install.sh: line 9: cd: ../../browser-extension: No such file or directory
```
```
npm ERR! enoent ENOENT: no such file or directory, open '/Users/nikunj/Documents/GitHub/requestly-selfhosted/common/analytics-vendors/dist/package.json'
```

### Repro steps

* run `find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +`, this deletes all node_modules folders for all modules.
* run `find . -name 'dist' -type d -prune -exec rm -rf '{}' +`, this deletes all dist folders for all modules.
* run `install.sh local`

### What Requestly tool were you using?

- [ ] Extension
- [ ] Desktop-App
- [ ] Android SDK
- [ ] Selenium

### Your Environment

Macos

### Requestly Version

latest

nkg447:
I am working on fixing this.

nkg447:
PR - https://github.com/requestly/requestly/pull/2289

Ronakkadhi:
Closing as stale — no activity for over a year.

As part of the 2026 repository restructure, we're closing dormant issues to focus on actively-tracked items. If this is still relevant, please re-open or file a fresh issue at the right home:

- 🌐 **HTTP Interceptor** (Chrome extension + web app + desktop proxy) → [requestly/interceptor](https://github.com/requestly/interceptor/issues/new/choose)
- 🚀 **Requestly API Client** → this repo

Thanks for taking the time to file the original report.

PR #2289, opened 2024-11-10
Fix install build scripts - issue - 2288

nkg447:
Fix for issue - https://github.com/requestly/requestly/issues/2288

Post this change install and build scripts runs without any error


app/package.json
```diff
@@ -58,7 +58,7 @@
     "@react-icons/all-files": "https://github.com/react-icons/react-icons/releases/download/v4.10.1/react-icons-all-files-4.10.1.tgz",
     "@reduxjs/toolkit": "^1.9.0",
     "@requestly/mock-server": "0.1.6",
-    "@requestly/requestly-core": "file:..",
+    "@requestly/requestly-core": "file:../dist",
```
nkg447, line 61: need to dist path as "module" key is defined in requestly-core's package.json

browser-extension/install.sh
```diff
@@ -2,12 +2,6 @@
 
 npm i -g web-ext
 
-echo -e "\n***** Installing analytics vendor's dependencies *****"
```
nkg447, line 5: removed this coz this is already done as part of browser-extention/common  preinstall scripts.

install.sh
```diff
@@ -32,4 +27,11 @@ rm -rf node_modules
 npm install
 cd ../..
 
+echo -e "\n***** Installing Browser Extension dependencies *****"
```
nkg447, line 30: browser-extention should be installed after analytics-vendor

SolsticeSpectrum:
```
index.js → dist/requestly-core.cjs.js, dist/requestly-core.esm.js...
[!] Error: Could not resolve './common/rule-processor/dist/' from index.js
Error: Could not resolve './common/rule-processor/dist/' from index.js
    at error (/home/user/requestly-selfhosted/node_modules/rollup/dist/shared/rollup.js:198:30)
    at ModuleLoader.handleResolveId (/home/user/requestly-selfhosted/node_modules/rollup/dist/shared/rollup.js:22287:24)
    at /home/user/requestly-selfhosted/node_modules/rollup/dist/shared/rollup.js:22250:26
```

SolsticeSpectrum:
Can you maybe try making github action in your fork? I maanged to compile it after a lot of headaches but I don't even remember what I did. One of the steps was `chmod -R 777 .`

nkg447:
To run the extension in local mode you need to build the extention locally. and import that in chrome.

SolsticeSpectrum:
> To run the extension in local mode you need to build the extention locally. and import that in chrome.

Well that does not work for me

install.sh
```diff
@@ -32,4 +27,11 @@ rm -rf node_modules
 npm install
 cd ../..
```
wrongsahil, line 28: Above installing analytics-vendors is not needed as it is already installed in the `browser-extension/install.sh`

browser-extension/install.sh
wrongsahil: This is required as preinstall script only builds the analytics vendor. It doesn't install the dependencies of the module

wrongsahil (changes requested):

browser-extension/install.sh
nkg447: the dependencies of the analytics-vendor module is installed in https://github.com/requestly/requestly/blob/master/install.sh

install.sh
nkg447: yeah, we can choose to keep the analytics-vendors `npm i` in either here or in browser-extention/install.sh

I have removed it from browser-extention/install.sh, do you want it other way round?

PR #2322, opened 2024-11-27
[ENGG-2459] feat: core engine for scripts in API client

nafees87n:


app/src/features/apiClient/APIClientManager/APIClientManager.ts
```diff
@@ -0,0 +1,26 @@
...
+  appMode: string,
+  request: any, //TODO: to be added with correct type
+  environmentManager: any,
+  signal?: AbortSignal
+): Promise<RQAPI.Response> => {
+  // Process request configuration with environment variables
+  const processedRequestConfig = environmentManager.renderVariables(request);
+
+  if (request.preRequestScript) {
+    executePrerequestScript(request.preRequestScript, processedRequestConfig, environmentManager);
```
wrongsahil, line 15: Isn't using worker as pre-scripts makes it async? 🤔

app/src/features/apiClient/APIClientManager/APIClientManager.ts
```diff
@@ -0,0 +1,46 @@
+import { makeRequest } from "../screens/apiClient/utils";
```
wrongsahil, line 1: Can we refractor the folder in the following folder structure. This will let us add more modules in future based on appMode (scripts modules for desktop app and extension can be different)


```
- features/apiClient
    - helpers
        - apiClientManager/
            - modules/
                - variables/
                - scripts/
             - index.ts 
    - screens
```

app/src/features/apiClient/APIClientManager/workerScripts.ts
```diff
@@ -0,0 +1,138 @@
...
+        console.error('Script error:', error);
+        throw error;
+      }
+      })();
+      `
+    );
+
+    scriptFunction(sandbox.rq).then(() => {
+      this.postMessage({
+        type: "HANDLE_ENVIRONMENT_CHANGES",
```
wrongsahil, line 52: This message name should be `SCRIPT_EXECUTED`.

app/src/features/apiClient/APIClientManager/APIClientUtils.ts
```diff
@@ -0,0 +1,138 @@
...
+    case "HANDLE_ENVIRONMENT_CHANGES": {
+      const updatedVariables = await handleEnvironmentChanges(environmentManager, payload);
+      cleanupWorker(worker);
+      resolve(updatedVariables);
+      break;
+    }
+
+    case "ERROR": {
+      cleanupWorker(worker);
+      resolve(null);
```
wrongsahil, line 67: In this case, it should result in an error instead of resolving

app/src/features/apiClient/APIClientManager/APIClientUtils.ts
```diff
@@ -0,0 +1,138 @@
...
+  const blob = new Blob([workerCode], { type: "application/javascript" });
+  return new Worker(URL.createObjectURL(blob));
+};
+
+const handleEnvironmentChanges = async (
+  environmentManager: any,
+  payload: {
+    variablesToSet: EnvironmentVariables;
+    variablesToRemove: string[];
+    currentVariables: EnvironmentVariables;
```
wrongsahil, line 27: Why are we passing `currentVariables`? Should set and remove be an upsert or remove operation (bulk update) and rest of the variables should stay as it is

app/src/features/apiClient/APIClientManager/workerScripts.ts
```diff
@@ -0,0 +1,138 @@
...
+        payload: {
+          currentVariables,
+          variablesToSet: variableSaverQueue,
+          variablesToRemove: variableRemoverQueue,
```
wrongsahil, line 56: Can we use this syntax here. (inspired from mongodb https://www.mongodb.com/docs/manual/reference/operator/update/set/#syntax
https://www.mongodb.com/docs/manual/reference/operator/update/unset/)
```
{
    "mutations": {
        "env": {
            "$set": {
                "key": "value"
            },
            "$unset": {
                "key": ""
            }
        },
        "collections": "// Support in Future"
    },
    "results": {
        // Will be adding test results here in Future
    }
}
```

app/src/features/apiClient/APIClientManager/workerScripts.ts
```diff
@@ -0,0 +1,138 @@
...
+      }
+      })();
+      `
+    );
+
+    scriptFunction(sandbox.rq).then(() => {
+      this.postMessage({
+        type: "HANDLE_ENVIRONMENT_CHANGES",
+        payload: {
+          currentVariables,
```
wrongsahil, line 54: I don't think we should trust currentVariables from here. We should only return the updated only from here and let the environmentManager merging logic handle non updated variables

app/src/features/apiClient/APIClientManager/workerScripts.ts
```diff
@@ -0,0 +1,138 @@
...
+        set: (key: string, value: any) => {
+          variableSaverQueue[key] = {
+            localValue: value,
+            syncValue: value,
+            type: typeof value,
+          };
+        },
+        get: (key: string) => {
+          const variable = currentVariables[key];
+          return variable?.localValue || variable?.syncValue;
```
wrongsahil, line 24: We should decide this logic in the `APIManager` only. APIManager should send only the final env values to here and let the variables processor handle the rest (type conversion, fallback values etc) in a single place.

```
{
    [key: string]: string|number...,
}
```

wrongsahil (changes requested):

app/src/features/apiClient/APIClientManager/workerScripts.ts
nafees87n: We do not have merging logic in environmentManager. setVariables always takes in the dump of all the variables

app/src/features/apiClient/helpers/APIClientManager/index.ts
```diff
@@ -0,0 +1,46 @@
+import { makeRequest } from "../../screens/apiClient/utils";
+import { RQAPI } from "../../types";
+import { executePrerequestScript, executePostresponseScript } from "./modules/scripts/utils";
```
wrongsahil, line 3: This shouldn't be in utils. Its should be the entry points and should be in `index.ts`. Please note this in refractoring

app/src/features/apiClient/APIClientManager/workerScripts.ts
wrongsahil: This is still not resolved yet.
Please make sure to incorporate this in refractoring

wrongsahil (changes requested):

wrongsahil:
Some bugs which i found
###IMP
- When some error in script occurs, the request should be aborted. Right now it keeps in loading
image
- When no environment is present, not even accessing the variables causes error even though request succeeds
- image



###Can be deferred
- When no environment present, case not handled (Causes error in console)

app/src/features/apiClient/types.ts
```diff
@@ -68,6 +68,7 @@ export namespace RQAPI {
       preRequest: string;
       postResponse: string;
     };
+    errorMessage?: string;
```
wrongsahil, line 71: This shouldn't be here as it is only required for pre-script and post-script. It is not related to entity

wrongsahil (changes requested):

app/src/features/apiClient/helpers/APIClientManager/modules/scripts/workerScripts.ts
```diff
@@ -0,0 +1,170 @@
...
+        remove: (key: string) => {
+          mutations.environment.$unset[key] = "";
+        },
+      },
+      variables: new Proxy(
+        {},
+        {
+          get: () => logDummyErrorMessage,
+        }
+      ),
```
wrongsahil, line 40: This might fail in case when the property which is accessed is not a function. 
eg. `rq.environment.name`

wrongsahil (changes requested):

wrongsahil (approved):

PR #2329, opened 2024-11-29
[ENGG-2413]Persist filters in traffic table

Aarushsr12:
- Filters added to the top filters in desktop app will be persisted after closing and reopening of the application
- Search bar will also persist the searched query


nsrCodes (approved):
LGTM 🔥

PR #2343, opened 2024-12-02
[ENG 2467] enhancement: breadcrumb support for sub collections

Ethical98:
Added Breadcrumb UI for sub collections


app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -175,13 +176,47 @@ export const convertFlatRecordsToNestedRecords = (records: RQAPI.Record[]) => {
...
+  formattedOptions = [
+    ...formattedOptions,
+    ...options.map((option) => ({
+      pathname: `${prefixOption.pathname}/${option.type}/${option.id}`,
+      label: option.name,
+      disabled: option.type === RQAPI.RecordType.COLLECTION,
+    })),
+  ];
+
+  formattedOptions.push({
```
wrongsahil, line 215: I don't think breadcrumbOptions needs to be pushed to db. They should be calculated at runtime

app/src/lib/design-system-v2/components/RQBreadcrumb/RQBreadcrumb.tsx
```diff
@@ -44,9 +49,13 @@ export const RQBreadcrumb: React.FC<Props> = ({
...
-  } & MatchedRoute["handle"]["breadcrumb"])[] = matchedRoutes.reduce((result, route) => {
-    return route.handle?.breadcrumb ? [...result, { ...route.handle.breadcrumb, pathname: route.pathname }] : result;
-  }, []);
+  } & MatchedRoute["handle"]["breadcrumb"])[] = !isEmpty(breadcrumbOptions)
+    ? breadcrumbOptions
+    : matchedRoutes.reduce((result, route) => {
+        return route.handle?.breadcrumb
+          ? [...result, { ...route.handle.breadcrumb, pathname: route.pathname }]
+          : result;
+      }, []);
```
wrongsahil, line 58: This doesn't look right. When breadcrumbOptions are present, it won't be rendering any other parent breadcrumb sections 🤔 ?

### Example
Expected - API > Collection1 > Collection2 > Request1
This -> Collection1 > Collection2 > Request1


We should only send extra prefix and rest of the things should be handled by RQBreadCrumb Component. i.e. 
adding the prefix at the (n-1) section. 
API > (here add prefix passed) > Request1
```
[
    { label: "Collection1", href: "/collection1"},
    { label: "Collection2", href: "/collection2"},
]
```

wrongsahil (changes requested):

app/src/features/apiClient/screens/apiClient/utils.ts
Ethical98: they are not pushed to db they are getting calculated only one time while rendering the sidebar UI then it is stored in apiClient context.

Ethical98:
Closing this PR will raise new PR according to new requirements.

PR #2354, opened 2024-12-03
ENGG-2465: Add shortcut for sending and saving requests in API Client

RuntimeTerror10:


nsrCodes (approved):
works, but again we don't have any events tracking if users are using shortcuts or button clicks. @echo-sg can you verify if that is important?

PR #2361, opened 2024-12-04
chore: refractor redux store

wrongsahil:
- Adds proper typescript types
- Adds autocomplete for globalActions for IDEs (when used in dispatch)

Solution for references not working in CaseReducers
https://github.com/microsoft/TypeScript/issues/59134


app/src/store/slices/global/constants.ts
```diff
@@ -1,15 +1,20 @@
 import APP_CONSTANTS from "config/constants";
 import { CONSTANTS as GLOBAL_CONSTANTS } from "@requestly/requestly-core";
-// SUB
-const appListJson = require("./sub/appsList.json");
+import { GlobalSliceState } from "./types";
 
-const INITIAL_STATE = {
+import appListJson from "../../initial-state/sub/appsList.json";
+
+console.log(appListJson);
```
RuntimeTerror10, line 7: Console log here, can be removed

RuntimeTerror10 (approved):

PR #2406, opened 2024-12-11
feat: Authorization tab

nsrCodes:
in both collection and requests


nsrCodes:
conflicted by https://github.com/requestly/requestly/pull/2409 but the UI there seems more stable on basic usage so going ahead with those changes with some manually copied patches from here

PR #2409, opened 2024-12-12
ENG-2463 -> Authorization Tab Support

Ethical98:


app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -184,3 +199,113 @@ export const convertFlatRecordsToNestedRecords = (records: RQAPI.Record[]) => {
...
 };
 export const getEmptyPair = (): KeyValuePair => ({ id: Math.random(), key: "", value: "", isEnabled: true });
+
+export const getEmptyAuthOptions = (): AUTH_OPTIONS => {
+  return {
+    [AUTHORIZATION_TYPES.NO_AUTH]: null,
+    [AUTHORIZATION_TYPES.API_KEY]: { key: "", value: "", addTo: "HEADER" },
+    [AUTHORIZATION_TYPES.BEARER_TOKEN]: { bearer: "" },
+    [AUTHORIZATION_TYPES.BASIC_AUTH]: { username: "", password: "" },
+  };
```
wrongsahil, line 209: This data looks unnecessary while saving to db. All the extra form data except the selected one should get stripped away. 

We can keep the previous state in the AuthorizationForm component only to keep the last know values while switching between different Auths. But the final onChange should only send the value for the selected auth type only

app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -184,3 +199,113 @@ export const convertFlatRecordsToNestedRecords = (records: RQAPI.Record[]) => {
...
+          AUTH_ENTRY_IDENTIFIER,
+          key || "",
+          value || ""
+        );
+      }
+      break;
+    }
+    default:
+      break;
+  }
```
wrongsahil, line 298: The addition of auth header should be a runtime logic. They shouldn't be added into the entry during the saving at all (although they are getting removed at the end in sanitise entry).

wrongsahil: This generation of headers should be somewhere over [here](https://github.com/requestly/requestly/blob/0ac721b84b5c2a31ac3ea0655ef88a30aff3f326/app/src/features/apiClient/helpers/APIClientManager/index.ts#L17) so that it doesn't effect the final entry which is saved in the db.

For showing in the Headers Panel, we should consume the same function which generated the headers and then should it separately which changing anything in the entry

wrongsahil (changes requested):

app/src/features/apiClient/screens/apiClient/components/clientView/APIClientView.tsx
```diff
@@ -352,6 +353,16 @@ const APIClientView: React.FC<Props> = ({ apiEntry, apiEntryDetails, notifyApiRe
     trackAPIRequestCancelled();
   }, []);
 
+  const handleAuthChange = useCallback(
+    (currentAuthType: AUTHORIZATION_TYPES, updatedKey: string, updatedValue: string, formValues: any) => {
+      setEntry((prevEntry) => {
+        const updatedEntry = updateAuthOptions(prevEntry, currentAuthType, updatedKey, updatedValue, formValues);
```
wrongsahil, line 359: Why are we passing `currentAuthType, updatedKey, updatedValue, formValues` all the way here? Shouldn't the AuthorizationView component handle the preparation of this and pass just the `AuthOptions` to the `handleAuthChange` method directly and this just becomes `entry.auth = value`

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -81,6 +83,24 @@ const processScripts = (item: any) => {
...
+    authOptionsArray.forEach((option) => {
+      auth[authType][option.key] = option.value;
+    });
+
+    return auth;
+  } catch (error) {
+    return {};
+  }
+};
+
```
wrongsahil, line 103: Please try to not mix features as possible in single PR. This should be in a separate PR that builds on top of this.

app/src/features/apiClient/helpers/APIClientManager/index.ts
```diff
@@ -15,6 +22,22 @@ export const executeAPIRequest = async (
...
   // Process request configuration with environment variables
+
+  if (!isEmpty(entry.auth)) {
+    const authType = entry.auth.currentAuthType;
+    const requestHeaders = entry.request.headers;
+    const queryParmas = entry.request.queryParams;
+    const addToQuery = authType === AUTHORIZATION_TYPES.API_KEY && entry.auth[authType].addTo === "QUERY";
+
+    if (addToQuery) {
+      entry.request.headers = deleteAuthOptions(requestHeaders);
```
wrongsahil, line 33: Why do we need to deleteAuthOptions first? We never added them during current execution cycle.

app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -184,3 +198,92 @@ export const convertFlatRecordsToNestedRecords = (records: RQAPI.Record[]) => {
...
+        updateDataInState(dataToUpdateCopy, AUTH_ENTRY_IDENTIFIER, key || "", value || "");
+      }
+      break;
+    }
+    default:
+      break;
+  }
+
+  return dataToUpdateCopy;
+};
```
wrongsahil, line 289: - Abstract this into a separate file `helpers/auth.ts`.

wrongsahil: Also we should update this method definition so that it can support showing the auth headers in the headers tab too. Something like this
```
function processAuth(auth: AuthOptions): {
  headers: { key: string; value: string }[]; // KeyValuePair // These are the auth headers that will be added to the request,
  params: { key: string; value: string }[]; // KeyValuePair, // These are the query parameters that will be added to the request
} {}

```

- Then we can consume these return values in `executeAPIRequest` and call a util method `updateRequestAuthHeaders` and `updateRequestQueryParams`
- And we can call the same method processAuth in the Headers Tab and use those values to render the new headers added by the auth

wrongsahil (changes requested):

app/src/features/apiClient/helpers/APIClientManager/index.ts
Ethical98: Case -> Added authorization api key to headers clicked on send then rendered request adds it to headers if user then changes from headers-> query params then we have to delete auth options from headers.

app/src/features/apiClient/helpers/auth.ts
```diff
@@ -0,0 +1,68 @@
...
+  const createAuthorizationKeyValuePair = (type: string, key: string, value: string): typeof newKeyValuePair => ({
+    id: Date.now(),
+    key,
+    value,
+    type,
+    isEnabled: true,
+  });
+
+  const updateDataInState = (data: KeyValuePair[], key: string, value: string) => {
+    const existingIndex = data.findIndex((option) => option.type === AUTH_ENTRY_IDENTIFIER);
```
wrongsahil, line 23: This looks redundant as the entries in the data corresponds to AuthEntries only. Instead we should check if the data contains an entry with same key

app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/AuthorizationView/authStaticData.ts
```diff
@@ -0,0 +1,112 @@
...
+    },
+  },
+  [AUTHORIZATION_TYPES.BASIC_AUTH]: {
+    formData: AUTHORIZATION_FORM_DATA[AUTHORIZATION_TYPES.BASIC_AUTH],
+    description: {
+      heading: "Basic Auth",
+      subHeading:
+        "API Key authentication allows you to send a key-value pair to the API for verification. Here’s how to set it up:",
+      note: "Store your key in a variable for enhanced security.",
+      externalLink: "https://app.requestly.io/",
```
wrongsahil, line 98: Link the learn more links to the docs

wrongsahil: Links are opening in the same tab right now. target='_blank' missing

wrongsahil:
## Test Cases
- [X] Bearer Token
- [X] Basic Auth
- [X] api_key
- [X] With Variable
- [X] Existing header already exist in header (dedup) (Right now sending multiple headers which work for now`    "authorization": "local, Bearer token testLocal",
`)

wrongsahil (approved):

app/src/features/apiClient/helpers/auth.ts
Ethical98: this was done to replicate the postman behaviour postman allows user to add the authorization header manually so if we go by the key then it will change the header added by user not by us using auhtorization tab.
[
Screenshot 2024-12-19 at 12 33 38 PM
](url)

Ethical98: there can only be one auth header so that's why we can reply on key

PR #2431, opened 2024-12-19
[ENGG-2646] UI Enhancement:  Show Auth Headers & Query Params

Ethical98:


app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/KeyValueTable/KeyValueTable.tsx
```diff
@@ -108,7 +116,7 @@ export const KeyValueTable: React.FC<KeyValueTableProps> = ({ data, setKeyValueP
         editable: true,
         onCell: (record: KeyValuePair) => ({
           record,
-          editable: true,
+          editable: record.type !== AUTH_ENTRY_IDENTIFIER,
```
wrongsahil, line 119: We should add `editable` prod in `KeyValueTableProps`. That will keep the auth logic outside without coupling that in this component

app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/RequestTabs/RequestTabs.tsx
```diff
@@ -58,18 +58,23 @@ const RequestTabs: React.FC<Props> = ({
...
-        ),
+        label: <LabelWithCount label="Query Params" count={queryParamsCount} />,
         children: (
-          <KeyValueTable
+          <KeyValueTableWIthAutoGeneratedValues
             data={requestEntry.request.queryParams}
             setKeyValuePairs={setRequestEntry}
             pairType={KeyValueFormType.QUERY_PARAMS}
             variables={variables}
+            autoGeneratedValues={autoGeneratedQueryParams}
```
wrongsahil, line 77: We should create a `RequestQueryParamsTab` component which is responsible for the following things
- Fetching processed auth params
- Rendering components
    - <KeyValuePair  data={processedAuthParams} editable={false} />
    - <KeyValuePair  data={entry.params} editable={true} />

Same for headers too. This will keep the logic in the specific components only for rendering the tab

wrongsahil: For rendering the label count, we can expose a handler which updates the count in the parent component

app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/KeyValueTable/KeyValueTableWIthAutoGeneratedValues.tsx
```diff
@@ -0,0 +1,44 @@
...
+  data,
+  autoGeneratedValues,
+  setKeyValuePairs,
+  variables,
+  pairType,
+}) => {
+  return (
+    <div className="keyvalue-table-wrapper">
+      {!isEmpty(autoGeneratedValues) && (
+        <KeyValueTable
```
wrongsahil, line 25: Need to add a tooltip on why these headers are disabled

app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/KeyValueTable/KeyValueTableWIthAutoGeneratedValues.tsx
```diff
@@ -0,0 +1,44 @@
...
+          data={autoGeneratedValues}
```
wrongsahil, line 26: Hiding the table autogenerated headers is missing right now

wrongsahil (changes requested):

app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/RequestTabs/RequestTabs.tsx
Ethical98: We can create different components for rendering the components but for label count it should stay in requestTabs component because count is shown on first mount for each tab if that logic is put in specific component then it will only change count when component tab is clicked. If you have any other approach please let me know.

app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/KeyValueTable/KeyValueTableWIthAutoGeneratedValues.tsx
Ethical98: DIdn't have the proper design for it. Will release in next version

PR #2411, opened 2024-12-12
feat: add auth tab to collection view

nsrCodes:

PR #2434, opened 2024-12-19
feat: add auth tab to collection view

nsrCodes:
cherry picks relevant changes from https://github.com/requestly/requestly/pull/2411 
attempts to bring those changes at par with what has recently been merged from the base PR https://github.com/requestly/requestly/pull/2409


nsrCodes:
There are glaring issues in the types of `AuthView` but I am neither touching those, nor adding `@ts-`directives when its consumed inside `CollectionView` here since they seem so obvious that I am assuming there was a reason for them not being pointed out/fixed in the [original PR](https://github.com/requestly/requestly/pull/2409) for this change

Ethical98 (approved):

PR #2412, opened 2024-12-12
feat: inherit from parent in auth

nsrCodes:

PR #2435, opened 2024-12-19
feat: inherit from parent

nsrCodes:
cherry picks (_with cautious modifications_) changes from https://github.com/requestly/requestly/pull/2412  

most of it needs to be redone based on the improvements later made and merged through https://github.com/requestly/requestly/pull/2409

### DIFF
--- a/app/src/features/apiClient/helpers/APIClientManager/index.ts
+++ b/app/src/features/apiClient/helpers/APIClientManager/index.ts
@@ -9,13 +9,23 @@ import { trackAPIRequestSent } from "modules/analytics/events/features/apiClient
 import { isEmpty } from "lodash";
 import { processAuthOptions, updateRequestWithAuthOptions } from "../auth";
 
+// import { AUTHORIZATION_TYPES } from "../../screens/apiClient/components/clientView/components/request/components/AuthorizationView/types";
 export const executeAPIRequest = async (
   appMode: string,
+  apiRecords: RQAPI.Record[],
   entry: RQAPI.Entry,
   environmentManager: any,
   signal?: AbortSignal,
   requestCollectionId?: string
 ): Promise<RQAPI.Entry | RQAPI.RequestErrorEntry> => {
+  // // get updated entry based on auth details
+  // const updatedEntry = updateAuthOptions(
+  //   apiRecords,
+  //   requestCollectionId,
+  //   entry,
+  //   entry.auth?.currentAuthType ?? AUTHORIZATION_TYPES.NO_AUTH
+  // ) as RQAPI.Entry;
+
   // Process request configuration with environment variables
 
   const updatedEntry = JSON.parse(JSON.stringify(entry)); //Deep Copy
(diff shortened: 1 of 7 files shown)


app/src/features/apiClient/helpers/APIClientManager/index.ts
```diff
@@ -7,26 +7,29 @@ import { renderTemplate } from "backend/environment/utils";
...
+import { processAuthForEntry, updateRequestWithAuthOptions } from "../auth";
 
 export const executeAPIRequest = async (
   appMode: string,
+  apiRecords: RQAPI.Record[],
   entry: RQAPI.Entry,
+  entryDetails: {
+    id: RQAPI.Record["id"];
+    collectionId: RQAPI.Record["collectionId"];
+  },
```
Ethical98, line 19: can this be prevented from passing? we can create this from entry.

app/src/features/apiClient/helpers/auth.ts
```diff
@@ -3,7 +3,55 @@ import { AUTHORIZATION_TYPES } from "../screens/apiClient/components/clientView/
...
+  },
+  allRecords: RQAPI.Record[]
+) {
+  const parentRecord = entryDetails.collectionId
+    ? allRecords.find((record) => record.id === entryDetails.collectionId)
+    : null;
+  if (!parentRecord) {
+    return null;
+  }
+  let parentAuthData = parentRecord.data.auth;
```
Ethical98, line 43: this can be prevented by consuming collection map once [PR](https://github.com/requestly/requestly/pull/2343) gets merged

app/src/features/apiClient/screens/apiClient/components/clientView/components/Collection/CollectionView.tsx
```diff
@@ -75,9 +75,12 @@ export const CollectionView = () => {
           auth: newAuthOptions,
         },
       };
-      return upsertApiRecord(user.details?.profile?.uid, record, teamId).then((result) => {
-        onSaveRecord(result.data);
-      });
+      return upsertApiRecord(user.details?.profile?.uid, record, teamId)
+        .then((result) => {
+          // fix-me: to verify new change are broadcasted to child entries that are open in tabs
```
Ethical98, line 80: is this done?

Ethical98 (changes requested):

app/src/features/apiClient/helpers/APIClientManager/index.ts
nsrCodes: this isn't part of `Entry`. It is part of the `Record`. Here's a reference to the [type defination for this](https://github.com/requestly/requestly/blob/master/app/src/features/apiClient/types.ts#L80-L131). 

It is clear that these are required to know what is the id of the record as well as the parent record

app/src/features/apiClient/helpers/auth.ts
nsrCodes: True and to be clear, this would only optimise line#38 here. 
IMO, I don't feel it would be correct to wait for some other PR to get merged as these changes have already had to be rewritten several times because of conflicting changes

app/src/features/apiClient/screens/apiClient/components/clientView/components/Collection/CollectionView.tsx
nsrCodes: As discussed, and was evident in the finally released [base PR](https://github.com/requestly/requestly/pull/2409), rendering correctness is no longer a concern and hence this comment is meant to be addressed only when we decide to fix that. So no, not done right now.

PR #2442, opened 2024-12-20
feat: auto complete plugin for variables in editor

nsrCodes:
file structure/organisation had to be opinionated as could not make this be driven from typescript interfaces. It is possible (_[example](https://github.com/val-town/codemirror-ts)_) but seemed like overkill for our usecase hence created a static list of details of the methods that are being added to `rq.` namespace in scripts 

Demo: 

https://github.com/user-attachments/assets/45f26ceb-4a0f-460e-afb9-73dedd5b90af


iostreamer-X (changes requested):
Hey @nsrCodes! Let's pause on this implementation for now.

Firstly, the code is good, but it's the fundamental behaviour that's of concern.
We didn't receive this requirement, rather we are the ones pushing for it, so there's no inherent rush to deliver.

Secondly, we are bringing ourselves at par with script editing experiences provided by other clients. And they have set the bar at full intellisense.
So we can't be delivering anything below that, as that might erode the trust on API client itself.

So, for now we will be pausing with this fuzzy search implementation and plan out for full intellisense.

wrongsahil:
> Hey @nsrCodes! Let's pause on this implementation for now.
> 
> Firstly, the code is good, but it's the fundamental behaviour that's of concern. We didn't receive this requirement, rather we are the ones pushing for it, so there's no inherent rush to deliver.
> 
> Secondly, we are bringing ourselves at par with script editing experiences provided by other clients. And they have set the bar at full intellisense. So we can't be delivering anything below that, as that might erode the trust on API client itself.
> 
> So, for now we will be pausing with this fuzzy search implementation and plan out for full intellisense.

We can keep the variables autocomplete though as it doesn't depend on intellisense

nafees87n (approved):

iostreamer-X (approved):

PR #2450, opened 2024-12-23
[ENGG-2546] Keyboard Shortcuts - UX improvement in API client

Aarushsr12:
- Use cmd+s to save & cmd+enter to hit the API request (on mac)
- Use ctrls+s to save & ctrl+enter to hit the API request (on windows/linux)
- Only Enter will be remove focus from input


app/src/features/apiClient/screens/apiClient/components/clientView/APIClientView.tsx
```diff
@@ -375,6 +375,21 @@ const APIClientView: React.FC<Props> = ({ apiEntry, apiEntryDetails, notifyApiRe
...
+          onSaveButtonClick();
+        } else if (evt.key.toLowerCase() === "enter") {
+          evt.preventDefault();
+          onSendButtonClick();
+        }
+      }
+    },
+    [onSaveButtonClick, onSendButtonClick]
+  );
+
```
RuntimeTerror10, line 392: `preventDefault()` is not required here as this is already handled in SingleLineEditors onKeyDown listener

RuntimeTerror10 (approved):

PR #2455, opened 2024-12-24
fix: unsaved changes reflected when a request is sent

RuntimeTerror10:


rohanmathur91 (changes requested):
Once request saved and any changes made then unsaved indication is not shown.

rohanmathur91 (approved):

PR #2461, opened 2024-12-26
ENG-2717 -> Add Inherit from auth in postman importer

Ethical98:
Changes:

1. Postman Default Authentication Cases:

- Postman does not export default authentication cases like `No Auth` for root collections or `Inherit from Parent` 
   for sub-collections.

- To handle these cases, the logic for `parentCollectionId` is added:
     - If no authentication options are defined during export and `parentCollectionId` is present, we will add 
     `INHERIT FROM PARENT`.
     - Otherwise, we will add `NO AUTH`.

2. Handling Default Authentication Values:

- When a new request or collection is created:
    - **Root Level**: The default value of authentication options is set to `NO AUTH`.
    - **Inside a Collection**: The default value of authentication options is set to `INHERIT FROM PARENT`.

- These default values are only displayed in the UI and are not stored in the database. To address this:

    - A change is made in the `execute API` method to handle default `authOptions` and append headers or query 
       parameters accordingly.

- Purpose: This change is implemented to mimic the Postman User Experience.


app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -82,21 +82,21 @@ const processScripts = (item: any) => {
...
-    const authType = POSTMAN_AUTH_TYPES_MAPPING[item.type] ?? AUTHORIZATION_TYPES.NO_AUTH;
-    auth.currentAuthType = authType;
-    auth[authType] = {};
-
-    const authOptionsArray = item[item?.type] || [];
-    authOptionsArray.forEach((option) => {
-      auth[authType][option.key] = option.value;
+    const currentAuthType =
+      POSTMAN_AUTH_TYPES_MAPPING[item.type] ??
+      (parentCollectionId ? AUTHORIZATION_TYPES.INHERIT : AUTHORIZATION_TYPES.NO_AUTH);
```
nsrCodes, line 89: this has edge cases for when imported auth type is `Inherit` and `no_auth` since they are not part of `POSTMAN_AUTH_TYPES_MAPPING`, please update that.

the two edge cases: 
1. if a request inside a collection (_has a parent Id_) and the request's auth inside postman is `no_auth`, then it will be set to `inherit` when importing here
2. if a top level collection has auth as `inherit` in postman it will get converted to `no_auth`  when imported to requestly

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -82,21 +82,21 @@ const processScripts = (item: any) => {
   return scripts;
 };
 
-const processAuthorizationOptions = (item = {}) => {
+const processAuthorizationOptions = (item: Record<string, any> = {}, parentCollectionId: string): RQAPI.AuthOptions => {
```
nsrCodes, line 85: so `parentCollectionId` should be optional here right?

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -82,21 +82,21 @@ const processScripts = (item: any) => {
...
+    const authOptions = item[item?.type] || [];
+    authOptions.forEach((option: Record<string, any>) => {
+      auth[currentAuthType][POSTMAN_FIELD_MAPPING.get(option.key)] = POSTMAN_FIELD_MAPPING.get(option.value);
     });
 
     return auth;
   } catch (error) {
-    return {};
+    return { currentAuthType: AUTHORIZATION_TYPES.NO_AUTH };
   }
```
nsrCodes, line 100: try catch seems redundant, can be removed

nsrCodes (changes requested):

nsrCodes (approved):
LGTM

PR #2501, opened 2025-01-06
[ENGG-2627] feat: add draft request flow in api client

rohanmathur91:


RuntimeTerror10:
Request is saved when renaming it. Also shows unsaved changes post saving

https://github.com/user-attachments/assets/530c598a-74ab-44cd-aabf-88ab36217e13

app/src/features/apiClient/contexts/apiClient.tsx
```diff
@@ -173,7 +174,7 @@ export const ApiClientProvider: React.FC<ApiClientProviderProps> = ({ children }
   );
 
   const onSaveRecord = useCallback(
-    (apiClientRecord: RQAPI.Record, openTabOnSave = true) => {
+    (apiClientRecord: RQAPI.Record, openTabOnSave = true, replaceTabOnSave = false) => {
```
RuntimeTerror10, line 177: Instead of having `replaceTabOnSave` and `openTabOnSave` flags. Can be keep a single param like `onSaveAction` which be either `replace | open` and based on the action we can call the Tab functions.

RuntimeTerror10: Considering this because in future we need handle the optional params and ordering of if checks as we keep on add this function's params

RuntimeTerror10 (approved):

PR #2506, opened 2025-01-07
[ENG-2786]: Dynamic bottom sheet height according to screen size

Ethical98:


app/src/componentsV2/BottomSheet/components/BottomSheetLayout/components/SplitPaneLayout/SplitPaneLayout.tsx
```diff
@@ -8,9 +8,10 @@ interface Props {
   bottomSheet: ReactNode;
   children: ReactNode;
   minSize?: number;
+  defaultSizing?: boolean;
```
RuntimeTerror10, line 11: Instead of passing default sizing as flag, we should should pass the sizes instead. As this layout is not only used in API Client but other screens as well ad depending on the screen we may change the default size when tab is opened

RuntimeTerror10 (approved):

PR #2526, opened 2025-01-08
[ENGG-2779] feat: add Requestly extension for Firefox

nafees87n:
## Summary of changes

1. refactored to use the `chrome.declarativeNetRequest` types in compile time only and never in runtime as some of the methods are `undefined` in Firefox.
2. used `cloneInto` to pass the object context from content script to page script. [Refer](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts#cloneinto)
3. Firefox does not use `documentLifecycle`. It has been handled as well.


browser-extension/config/configs/browser/firefox.json
```diff
@@ -1,4 +1,4 @@
 {
   "storageType": "local",
-  "contextMenuContexts": ["all"]
+  "contextMenuContexts": ["browser_action"]
```
nafees87n, line 3: Was used in MV2 but not required now. Not cleaning it up now.

browser-extension/mv3/package.json
```diff
@@ -43,6 +44,7 @@
     "rollup": "^2.79.1",
     "rollup-plugin-copy": "^3.4.0",
     "tslib": "^2.4.0",
-    "typescript": "^4.9.5"
+    "typescript": "^4.9.5",
+    "web-ext": "^8.3.0"
```
wrongsahil, line 48: Can you remove web-ext from the install.sh script

wrongsahil (approved):

PR #2533, opened 2025-01-09
[ENGG-2735] fix: current values get cleared on switching workspace

nafees87n:
# Changes
- Changed redux structure to store the environments mapped to `ownerId`

### DIFF
--- a/app/src/store/features/variables/reducers.ts
+++ b/app/src/store/features/variables/reducers.ts
@@ -15,9 +15,10 @@ const addNewEnvironment = (
   action: PayloadAction<{
     id: string;
     name: string;
+    ownerId: string;
   }>
 ) => {
-  state.environments[action.payload.id] = {
+  state.environments[action.payload.ownerId][action.payload.id] = {
     id: action.payload.id,
     variables: {},
     name: action.payload.name,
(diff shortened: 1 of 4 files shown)


app/src/store/features/variables/reducers.ts
```diff
@@ -15,9 +15,10 @@ const addNewEnvironment = (
   action: PayloadAction<{
     id: string;
     name: string;
+    ownerId: string;
   }>
 ) => {
-  state.environments[action.payload.id] = {
+  state.environments[action.payload.ownerId][action.payload.id] = {
```
wrongsahil, line 21: @nafees87n  why `state.environments.<ownerId>.<environmentId>` ? why not `state.environments..<environmentId>` directly?

nafees87n: Changed the data structure [here](https://github.com/requestly/requestly/pull/2533/files#diff-43dcf15560a89a7da8a241595dff94c6f78547275c538f81b263aee10e909930R5-R7)

All the user's environment will be cached in redux lazily as user keeps on switching across workspaces

RuntimeTerror10:
The `getEnvironmentsByOwnerId ` function by name it seems like it gives env map for a given ownerId (passing a specific ownerId to the function), but it returns the env map of active owner.

Can we write it like this instead?

```js 
const activeOwnerEnvironments = useMemo(() => allEnvironmentData?.[ownerId] ?? {}, [allEnvironmentData, ownerId]);
```

wdyt?

PR #2540, opened 2025-01-10
[ENGG-2800]: Refactor KeyValueTable component

RuntimeTerror10:
The component is currently highly coupled with API requestEntry state

* The table should only accept the `pairs`, `onChange` handler and `variables`

*  Let the respective parent component of KeyValueTable handle the request entry update

* Avoid passing `pairType`  prop,  parent will handle it.

### DIFF
--- a/app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/KeyValueTable/KeyValueTable.tsx
+++ b/app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/KeyValueTable/KeyValueTable.tsx
@@ -158,7 +158,7 @@ export const KeyValueTable: React.FC<KeyValueTableProps> = ({ data, variables, o
       rowKey="id"
       columns={columns as ColumnTypes}
       data={memoizedData}
-      locale={{ emptyText: `No query params found` }}
+      locale={{ emptyText: `No entries found` }}
       components={{
         body: {
           row: EditableRow,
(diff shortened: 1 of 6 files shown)


app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/KeyValueTable/KeyValueTable.tsx
```diff
@@ -178,7 +160,7 @@ export const KeyValueTable: React.FC<KeyValueTableProps> = ({ data, setKeyValueP
       rowKey="id"
       columns={columns as ColumnTypes}
       data={data}
-      locale={{ emptyText: `No ${pairType} found` }}
+      locale={{ emptyText: `No query params found` }}
```
iostreamer-X, line 163: Shouldn't this be something generic?

app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/KeyValueTable/KeyValueTableRow.tsx
```diff
@@ -50,8 +46,8 @@ export const EditableCell: React.FC<React.PropsWithChildren<EditableCellProps>>
     try {
       const values = await form.validateFields();
       handleUpdatePair({ ...record, ...values });
-    } catch (errInfo) {
-      Logger.log(pairtype, " KeyValueTable: Save failed:", errInfo);
+    } catch (error) {
+      console.error("Error saving key-value pair", error);
```
iostreamer-X, line 50: Are we losing anything by not using common Logger?

RuntimeTerror10: `Logger` is our internal utility which helps us debugging in prod, but this not captured by Sentry because this is disabled by default. That's why using `console.error` for now

PR #2542, opened 2025-01-10
feat apiClient: bruno importer

nsrCodes:


app/src/features/apiClient/contexts/apiClient.tsx
```diff
@@ -291,7 +291,7 @@ export const ApiClientProvider: React.FC<ApiClientProviderProps> = ({ children }
           setIsRecordBeingCreated(recordType);
           trackCreateEnvironmentClicked(analyticEventSource);
           return addNewEnvironment("New Environment")
-            .then((newEnvironment: { id: string; name: string }) => {
+            .then((newEnvironment: { id: string; name: string; isGlobal: boolean }) => {
```
Ethical98, line 305: whats the need for this ? like we are already maintaining the id==="global" for global environment? Changing this will result in changes in other importer logic also and currently we don't use any isGlobal check app wide so i don't think its necessary to make this change. If you have any other thoughts on this we can discuss

Ethical98 (changes requested):

app/src/features/apiClient/screens/BrunoImporter/utils.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+  return {
+    id,
+    name,
+    collectionId: parentCollectionId || "",
+    deleted: false,
+    type: RQAPI.RecordType.COLLECTION,
+    data: {
+      variables,
+      auth: processAuthorizationOptions(auth),
```
Ethical98, line 150: parentCollectionId is missing here

app/src/features/apiClient/screens/BrunoImporter/utils.tsx
```diff
@@ -0,0 +1,232 @@
...
+    data: {
+      request: {
+        url: request.url,
+        method: (request.method || "GET") as RequestMethod,
+        queryParams: processParams(request.params),
+        headers: processParams(request.headers),
+        body: requestBody,
+        contentType,
+      },
+      auth: processAuthorizationOptions(request.auth),
```
Ethical98, line 108: pass parentCollectionId here

app/src/features/apiClient/screens/BrunoImporter/types.ts
```diff
@@ -0,0 +1,100 @@
...
+    | AuthType<"digest", { username: string; password: string }>
+    | AuthType<"none", never>;
+
+  export interface Variable extends FieldValue {
+    local: boolean;
+    secret?: boolean;
+    type?: string;
+  }
+
+  export type Header = FieldValue;
```
Ethical98, line 81: header also requires "type" property

Ethical98 (changes requested):

app/src/features/apiClient/contexts/apiClient.tsx
nsrCodes: I have not made any changes here, just fixed a ts error by updating the explicit type user here to match the actual return type of the `addNewEnvironment` function - [Ref](https://github.com/requestly/requestly/blob/master/app/src/backend/environment/hooks/useEnvironmentManager.ts#L79-L83)

so I am not sure how this raises any questions about the data structure for global environments.

app/src/features/apiClient/screens/BrunoImporter/types.ts
nsrCodes: do you have an export that has this? cause I could not see this any exports I created.
And I used their export code as a reference and it is not mentioned there either - [Ref](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/utils/collections/index.js#L224-L247)

app/src/features/apiClient/screens/BrunoImporter/types.ts
```diff
@@ -0,0 +1,100 @@
...
+      name: string;
+      variables: Variable[];
+    }>;
+    brunoConfig?: {
+      version: string;
+      name: string;
+      type: string;
+      ignore?: string[];
+    };
+    root?: RootData;
```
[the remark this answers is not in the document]
nsrCodes: There are definitely similarities between the two types. `RootCollection` essentially extends the `Item[type="folder"]` type, but IMO they are independent entities with important differences in the DS (to the point of having explicitly different names (the internal items are called `folder` but there is no name in their code for the top level collection)). Here's an example export:


Screenshot 2025-01-28 at 1 15 12 AM


So it did not seem like there is a dependence in the types and their usage to lead to a `extends Item` kind of approach for `RootCollection`. 

But I am okay with doing so if you still believe that should be done

app/src/features/apiClient/screens/BrunoImporter/types.ts
Ethical98: i am saying type is optional but is there in case for auth headers, no issues we can skip it for now currently we dont show auth headers on UI so it wont cause any issue.

app/src/features/apiClient/screens/BrunoImporter/utils.tsx
```diff
@@ -0,0 +1,224 @@
...
+import { AUTHORIZATION_TYPES } from "features/apiClient/screens/apiClient/components/clientView/components/request/components/AuthorizationView/types";
+import { Bruno } from "./types";
+
+export const processBrunoScripts = (request: Bruno.Request) => {
+  const scripts = {
+    preRequest: request?.script?.req || "",
+    postResponse: request?.script?.res || "",
+  };
+  return scripts;
+};
```
Ethical98, line 13: replacing `bru` to `rq` is missing here while importing scripts

Ethical98:
Migration of bruno methods (getCollectionVar,setCollectionVar, etc) to our in house methods is still pending @nsrCodes. 
Approving the PR.

Ethical98 (approved):

PR #2553, opened 2025-01-13
[ENGG-2788]refactor: Code editor component

Aarushsr12:
PR focuses on:

- Codemirror Cursor bug fixed [when typing fast cursor sets at the beginning of editor]
- Refactored Editor component for API client only
- Prettification in RequestBody & Response Body


app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -0,0 +1,361 @@
...
+const Editor: React.FC<EditorProps> = ({
+  value,
+  language,
+  isReadOnly = false,
+  height = 225,
+  isResizable = false,
+  hideCharacterCount = false,
+  handleChange = () => {},
+  toolbarOptions,
+  id = "",
```
iostreamer-X, line 57: Hey, just wanted to understand the repercussions of this. If I have two editor instances on the same page, and I didn't provide `id` to any of them. Would that mean two dom nodes with same id?

Side note: I don't see this id being used anywhere.

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+  const editorRefCallback = (editor: ReactCodeMirrorRef) => {
+    if (!editorRef.current && editor?.editor && editor?.state && editor?.view) {
+      editorRef.current = editor;
+      setIsEditorInitialized(true);
+    }
+  };
+
+  const updateContent = useCallback((prettifiedCode: string) => {
+    if (editorRef.current?.view) {
```
iostreamer-X, line 126: Code nitpick since it makes it a little easier:

```ts
() => {
  const view = editorRef.current?.view;
  if(!view) {
    return;
  }
  
  // Use view normally here
}
```

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -0,0 +1,361 @@
...
+  }, [language]);
+
+  const editorRefCallback = (editor: ReactCodeMirrorRef) => {
+    if (!editorRef.current && editor?.editor && editor?.state && editor?.view) {
+      editorRef.current = editor;
+      setIsEditorInitialized(true);
+    }
+  };
+
+  const updateContent = useCallback((prettifiedCode: string) => {
```
iostreamer-X, line 125: The signature makes it look like this only updates prettified code but the code looks like it can update the entirety of content.

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -0,0 +1,361 @@
...
+        updateContent(prettified.code);
+      }
+    }
+  }, [config?.enablePrettify, language, value, updateContent]);
+
+  useEffect(() => {
+    if (isEditorInitialized && !isDefaultPrettificationDone.current) {
+      applyPrettification();
+      isDefaultPrettificationDone.current = true;
+    } else if (!prettifyOnInit) {
```
iostreamer-X, line 148: Can't we reach this branch if editor is not initialized? And then we might attempt to prettify even in that scenario?

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -0,0 +1,361 @@
...
+      //on render
+      applyPrettification();
+    }
+  }, [isEditorInitialized, isDefaultPrettificationDone, applyPrettification, prettifyOnInit]);
+
+  const handleEditorClose = useCallback(
+    (id: string) => {
+      // @ts-expect-error
```
iostreamer-X, line 156: Can you please add some context here? Why expect error?

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -0,0 +1,361 @@
...
+              view.dom.dispatchEvent(event);
+              return true;
+            },
+          },
+        ])
+      ),
+    []
+  );
+
+  return isFullScreen ? (
```
iostreamer-X, line 201: Please create either smaller components in the same file, or local variables in this function. And then use them in conditional branches.

iostreamer-X (changes requested):

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
Aarushsr12: 'id' prop contains the script id passed from custom script row

- it is being used in EditorToastContainer to give warning in insert script rule 
- `EditorToastContainer` component does not effect editor v2, we can keep this or either remove it
- we can also remove this from here and define it in custom script row component

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
Aarushsr12: This is fixed

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
Aarushsr12:
```
// @ts-expect-error
dispatch(globalActions.removeToastForEditor({ id }));
```

- `removeToastForEditor` reducer function is in `editorToastAction.js` , so to suppress that error from being reported in `Editor.tsx` file 
- To remove this need to changes editorToastAction to `editorToastAction.ts`

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
Aarushsr12: Defined local variables in function

iostreamer-X (approved):

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -0,0 +1,338 @@
...
+  const [isFullScreen, setIsFullScreen] = useState(false);
+  const [hoveredVariable, setHoveredVariable] = useState(null);
+  const [popupPosition, setPopupPosition] = useState({ x: 0, y: 0 });
+  const isFullScreenModeOnboardingCompleted = useSelector(getIsCodeEditorFullScreenModeOnboardingCompleted);
+  const [isEditorInitialized, setIsEditorInitialized] = useState(false);
+  const allEditorToast = useSelector(getAllEditorToast);
+  const toastOverlay = useMemo(() => allEditorToast[scriptId], [allEditorToast, scriptId]); // todo: rename
+  const [isCodePrettified, setIsCodePrettified] = useState(prettifyOnInit);
+  const isDefaultPrettificationDone = useRef(false);
+  const [editorContent, setEditorContent] = useState(value);
```
RuntimeTerror10, line 76: This now creates 2 source of truths for editor component. some effects use `value` prop and `editorContent`. we should have a single source of truth.

RuntimeTerror10, line 76: This now creates 2 source of truths for editor component. some effects use `value` prop and `editorContent`. we should have a single source of truth.

RuntimeTerror10 (changes requested):

nsrCodes (approved):
communicated suggestions over slack, looks good for now

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
Aarushsr12: This is fixed, removed the second state, considering only the value to be consistent source of truth

RuntimeTerror10 (approved):

PR #2557, opened 2025-01-14
[ENGG-2814] fix: query params lost in history request

rohanmathur91:


app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -305,7 +305,7 @@ export const syncQueryParams = (
       const updatedUrl = queryParamsToURLString(queryParams, url);
 
       // Dont sync if URL is same
-      if (updatedUrl !== url) {
+      if (updatedUrl !== url || !isEmpty(updatedQueryParams)) {
```
Ethical98, line 308: i think this will be a better check `if (updatedUrl !== url || updatedQueryParams.length !== queryParams.length)` as it will prevent syncing to run every time. It will run only in case of history or whenever there is a mismatch in length of queryParams.

Ethical98 (changes requested):

app/src/features/apiClient/screens/apiClient/utils.ts
rohanmathur91: It wont sync in runtime, since on first render it will sync the params

Ethical98: Syncing is necessary to ensure backward compatibility for requests. Once the initial sync is completed and the feature goes live, there is no need for syncing during the first render. If syncing occurs on every render, the logic will remove duplicate query parameters which is supported in Postman, which is why I want to avoid it.

However, I just realized that my current approach will not solve this as this will remove the duplicate queryParams on history view. We should not sync queryParams after the initial sync, as this would lead to the loss of duplicate parameters.

Suggestion:
We can add this 
` isEmpty(apiEntry.request.queryParams) ? QueryParamSyncType.TABLE : QueryParamSyncType.SYNC`
in APIClientView.tsx file at

`useEffect(() => {
    if (apiEntry) {
      setEntry({
        ...apiEntry,
        request: { ...apiEntry.request, ...syncQueryParams(apiEntry.request.queryParams, apiEntry.request.url) },
      });
      setRequestName("");
    }

    animationTimerRef.current = setTimeout(() => setIsAnimating(false), 800);

    return () => {
      clearTimeout(animationTimerRef.current);
    };
  }, [apiEntry]);
`

This will only sync the queryTable when queryParams are empty basically re add all the queryParam from the URL in this approach we'll not lose any duplicate queryParams.

Ethical98 (approved):

PR #2580, opened 2025-01-21
[ENG-2848]: Variable data loss & reordering fix

Ethical98:
1. Add id to maintain order for variables.
2. Fix data loss on switching variable type from number to secret.


app/src/features/apiClient/screens/environment/components/VariablesList/components/customTableRow/CustomTableRow.tsx
```diff
@@ -183,6 +190,15 @@ export const EditableCell: React.FC<EditableCellProps> = ({
...
 
+  useEffect(() => {
+    if (!isEmpty(record)) {
+      setPreviousValues((prev) => ({
+        ...prev,
+        [record.type]: { syncValue: record.syncValue, localValue: record.localValue },
+      }));
+    }
+  }, [record]);
+
```
RuntimeTerror10, line 201: This effect will set the state on any value change as well. We can avoid this unnecessary state updates and only set the previous values when record type is changed.

RuntimeTerror10: Also instead of a side effect can we have this logic with in the variable type change handler function?

app/src/backend/environment/hooks/useEnvironmentManager.ts
RuntimeTerror10, line 1: We need to manage Ids not only when creating variables, but also when deleting. example, I have 3 variables with ids 0,1,2 respectively. I delete the second variable. When I add new variable it will be assigned an id with value 2. Then any change in the new variable will also affect the other variable with id 2

app/src/features/apiClient/screens/environment/components/VariablesList/components/customTableRow/CustomTableRow.tsx
```diff
@@ -66,38 +68,43 @@ export const EditableCell: React.FC<EditableCellProps> = ({
...
+          currentValue.syncValue = String(previousValue.syncValue);
+          currentValue.localValue = String(previousValue.localValue);
+          break;
+        default:
           break;
       }
 
-      handleVariableChange({ ...record, type: value, ...defaultValues }, "type");
+      handleVariableChange({ ...record, type: newType, ...currentValue }, "type");
     },
```
RuntimeTerror10, line 106: This function has become very complex with nested switches. Can we have helper functions to convert values?

RuntimeTerror10 (changes requested):

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -254,13 +265,20 @@ export const processPostmanVariablesData = (
...
-    };
-    return acc;
-  }, {});
+  const variables = fileContent.variable.reduce(
+    (acc: Record<string, EnvironmentVariableValue>, variable: any, index: number) => {
+      acc[variable.key] = {
+        id: index,
+        syncValue: variable.value,
+        type:
+          variable.type === EnvironmentVariableType.String
```
RuntimeTerror10, line 274: `variable.type === EnvironmentVariableType.Secret`

app/src/features/apiClient/screens/environment/components/VariablesList/VariablesList.tsx
```diff
@@ -105,12 +106,15 @@ export const VariablesList: React.FC<VariablesListProps> = ({ searchValue = "",
 
   const handleDeleteVariable = useCallback(
     async (id: number) => {
-      const newData = id ? dataSource.filter((item) => item.id !== id) : dataSource.slice(0, -1);
+      const newData = id
+        ? dataSource.filter((item) => item.id !== id).map((record, index) => ({ ...record, id: index }))
+        : dataSource.slice(0, -1);
```
RuntimeTerror10, line 111: Lets return if no id is present

RuntimeTerror10 (changes requested):

RuntimeTerror10 (approved):

PR #2584, opened 2025-01-21
copy: mention that invites can also be for a group of emails

nsrCodes:


nsrCodes:
This was supposed to be a draft PR, closing for now.

PR #2600, opened 2025-01-26
ENGG-2882: Fix rule editor crashes for document language other than `en`

RuntimeTerror10:


app/src/hooks/useAppLanguage.ts
```diff
@@ -0,0 +1,26 @@
+import { useEffect } from "react";
+import { useSelector, useDispatch } from "react-redux";
+import { getAppLanguage } from "store/selectors";
+import { globalActions } from "store/slices/global/slice";
+
+export const useAppLanguage = () => {
```
rohanmathur91, line 6: Rename it to `useAppLanguageObserver`

app/src/hooks/useAppLanguage.ts
```diff
@@ -0,0 +1,26 @@
...
+  const dispatch = useDispatch();
+  const currentLanguage = useSelector(getAppLanguage);
+
+  useEffect(() => {
+    const observer = new MutationObserver(() => {
+      const newLang = document.documentElement.lang;
```
rohanmathur91, line 12: Rare case, but check if document exists, before getting `documentElement`

app/src/hooks/useAppLanguage.ts
```diff
@@ -0,0 +1,26 @@
...
+      if (newLang !== currentLanguage) {
```
rohanmathur91, line 13: No need of this check, if `newLang === currentLanguage` there will be no update after the dispatch, since the state is same, this will also remove the `currentLanguage` from dependency array.

app/src/lib/design-system-v2/components/RQButton/RQButton.tsx
```diff
@@ -92,7 +94,9 @@ interface RQCustomButton
   extends React.ForwardRefExoticComponent<RQButtonProps & React.RefAttributes<HTMLButtonElement>> {}
 
 export const RQButton: RQCustomButton = React.forwardRef<HTMLButtonElement, RQButtonProps>(function Button(props, ref) {
-  if (props.hotKey) {
+  const appLanguage = useSelector(getAppLanguage);
+
+  if (props.hotKey && appLanguage === "en") {
```
rohanmathur91, line 99: Add a comment, why we added the language condition here

rohanmathur91 (changes requested):

rohanmathur91 (approved):

PR #2620, opened 2025-01-30
[ENGG-2780] feat: Safari extension

nafees87n:
Creates extension build script for Safari. Currently only API client will be supported on Safari


app/src/actions/ExtensionActions.js
```diff
@@ -18,6 +18,10 @@ export function isExtensionInstalled() {
   return !!getExtensionVersion();
 }
 
+export const isSafariExtension = () => {
+  return !!document.documentElement.getAttribute("rq-is-safari");
```
wrongsahil, line 22: Can't we use navigator api or ua-parser directly here?

PR #2625, opened 2025-01-31
[ENGG-2837] fix: dynamic response rule returns null response on error

nafees87n:
## Summary of changes
In case of javascript errors in dynamic response rule, the fetch/XHR API shouldn't fail. Instead only the modification should fail and the original server response should returned.

This PR fixes and returns the original response in case of error:
1. XHR: The implementation was buggy, original response was not being returned.
2. Fetch: The response body was already read in the [page script](https://github.com/requestly/requestly/pull/2625/files#diff-397ea582548fc426ca8a0e2d6612d9a8902a31fba0db950d6e82f0d69129c9a1R171-R172), hence made a clone for the fetchedResponse to be returned.


wrongsahil (commented):
Can you add some details here. Why are we doing this and how does it resolve it

browser-extension/mv3/src/page-scripts/ajaxRequestInterceptor/xhr.js
```diff
@@ -121,7 +121,7 @@ export const initXhrInterceptor = (debug) => {
         }
 
         if (typeof customResponse === "undefined") {
-          return;
+          customResponse = this.response;
```
wrongsahil, line 124: I don't think this is the only thing that will solve this issue. if `getFunctionFromCode` throws an error, this if won't event capture it

wrongsahil (changes requested):

PR #2627, opened 2025-01-31
[ENGG-2650] feat: redirect to last visited feature

Aarushsr12:
- Redirect the user to last visited feature
- For both loggedIn/Not loggedIn users redirected to last visited feature
- It works for desktop application too

https://github.com/user-attachments/assets/7c84953a-74e2-4e06-9b60-e073fb8463bb


app/src/hooks/usePathFeatureTracker.ts
```diff
@@ -0,0 +1,25 @@
...
+  const lastFeaturePath = useSelector(getLastVisitedPath);
+
+  useEffect(() => {
+    if (location.pathname === "/" && lastFeaturePath && lastFeaturePath !== "/") {
+      navigate(lastFeaturePath, { replace: true });
+    }
+  }, [lastFeaturePath, location.pathname, navigate]);
+
+  useEffect(() => {
+    const featurePath = "/" + location.pathname.split("/")[1];
```
nsrCodes, line 20:
```suggestion
    const featurePath = "/" + location.pathname.split("/")?.[1] ?? "";
```

app/src/layouts/AppLayout/index.tsx
```diff
@@ -40,6 +41,7 @@ const App: React.FC = () => {
     growthbook.loadFeatures({ autoRefresh: true });
   }, []);
 
+  useLastFeatureTracker();
```
nsrCodes, line 44: This places it above the [`RootComponent`](https://github.com/requestly/requestly/blob/master/app/src/components/redirects/RootComponent.tsx#L17) that has its own sets of redirects

That component is where the [`redirectToRoot`](https://github.com/requestly/requestly/blob/master/app/src/utils/RedirectionUtils.js) expects to land. Just make sure that does not break (I have done basic testing for this, so far it works well)

nsrCodes (changes requested):
I believe there is a inclusion/exclusion list that is required here as we would not want to remember every possible route.

For example: this lead to a bug when if I got to an invite and then continue my browsing and then try to go via the extension, the app breaks

<video src="https://github.com/user-attachments/assets/5ded624c-a327-4e82-a3ee-d1236ab2abca" />

I can see similar cases for draft sessions, pricing page redirects, magic link (that route where we ask the user to enter their email input), etc.

PR #2647, opened 2025-02-05
Fix: Collection variables crash

Ethical98:


app/src/features/apiClient/screens/environment/components/VariablesListHeader/VariablesListHeader.tsx
nafees87n, line 1: Lets make this optional in the component's prop type as well

nafees87n (approved):

PR #2649, opened 2025-02-06
[ENGG-2855] chore: move svgs to public folder & update imports

Aarushsr12:
- moved svgs to public folder & removed imports to improve app performance


wrongsahil:
Looks good. LGTM

PR #2651, opened 2025-02-06
[ENGG-2957] fix: flicker when saving environment variables

nafees87n:


RuntimeTerror10:
Instead of adding an effect to sync the `pendingVariables` state with the `variables`, can we update the state in `handleSaveVariables` function.

RuntimeTerror10 (approved):

PR #2661, opened 2025-02-10
[ENGG-2994] chore: fix typescript types

rohanmathur91:
Changes in:
- src/backend/mocks
- src/backend/sessionRecording
- src/backend/workspace
- src/features/incentivization
- src/features/mocks
- src/hooks
- src/layouts/AppLayout
- src/layouts/DashboardLayout
- src/routes/index.tsx:39


RuntimeTerror10 (approved):

app/src/backend/workspace/index.ts
nafees87n, line 1: Can we write both the functions using await syntax instead of then?
Moreover there's inconsistency between return type in case of error. catche returns array while the funtion returns an object
```js
try{
    const res = await getInvites({ email, domain });
    return res?.data;
  }catch(e){
    console.log("error", e);
    return [];
  }
```

app/src/features/mocks/modals/ExportMocksModalWrapper/utils.ts
```diff
@@ -27,7 +27,7 @@ export const prepareMocksToExport = (mocks: RQMockSchema[]) => {
   });
 
   const fileName =
-    mocks.length === 1 ? `${mocks[0].name}` ?? "" : `requestly_mocks_export_${getFormattedDate("DD_MM_YYYY")}`;
+    mocks.length === 1 ? `${mocks[0].name}` || "" : `requestly_mocks_export_${getFormattedDate("DD_MM_YYYY")}`;
```
nafees87n, line 30:
```suggestion
    mocks.length === 1 ? `${mocks[0].name ?? ""}` : `requestly_mocks_export_${getFormattedDate("DD_MM_YYYY")}`;
```

left string will always be truthy in case of the OR operator as well

app/src/hooks/useGrowthbookIntegration.ts
```diff
@@ -13,7 +13,7 @@ const useGrowthBookIntegration = () => {
   const userAttributes = useSelector(getUserAttributes);
 
   const usePrevious = (value: any) => {
-    const ref = useRef();
+    const ref = useRef<unknown>(undefined);
```
nafees87n, line 16: Is this really required?

nafees87n (changes requested):
`app/src/backend/workspace/index.ts` some functions have type inconsistencies

app/src/hooks/useGrowthbookIntegration.ts
rohanmathur91: Needed initial value, type can be ignore, updated

nafees87n (approved):

PR #2662, opened 2025-02-11
fix: typescript config - compilation error

nafees87n:
Typescript was not compiling because `moduleResolution` was set as `bundler` in [this commit](https://github.com/requestly/requestly/blob/cf51d760185ab553f66f150d2f537e04d52a8a7b/app/tsconfig.json).

The above change was done to import typing from `@requestly/shared` module.

This PR 
1. Resets `moduleResolution` to `node`
2. Fixes shared module's `typesVersions` to import types from subpaths of `@requestly/shared`

### DIFF
--- a/app/tsconfig.json
+++ b/app/tsconfig.json
@@ -12,7 +12,7 @@
     "noUnusedLocals": true,
     "noImplicitAny": true,
     "noImplicitReturns": false,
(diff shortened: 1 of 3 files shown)


wrongsahil:
Can we close this now? @nafees87n

PR #2665, opened 2025-02-11
[ENGG-2650] refactor: move last feature path tracking to DashboardLayout

Aarushsr12:


app/src/routes/homeRoutes.tsx
```diff
@@ -0,0 +1,14 @@
+import RootComponent from "components/redirects/RootComponent";
```
wrongsahil, line 1: - app/src/features/home
    - routes.tsx


We can later on move other components of HomePage here too from `components` folder. For now this will be a empty folder

app/src/routes/homeRoutes.tsx
```diff
@@ -0,0 +1,14 @@
...
+import PATHS from "config/constants/sub/paths";
+import { RouteObject } from "react-router-dom";
+
+export const homeRoutes: RouteObject[] = [
+  {
+    path: PATHS.ROOT,
+    element: <RootComponent />,
```
wrongsahil, line 8: I don't think we need both RootComponent and `useRedirectToLastFeature`. Only `useRedirectToLastFeature` will suffice. Can we remove the RootComponent and move its logic to `useRedirectToLastFeature` only

wrongsahil (changes requested):

app/src/routes/homeRoutes.tsx
wrongsahil: This is independent of current changes. We can do this later too

app/src/features/home/route.tsx
wrongsahil, line 1: routes.tsx

wrongsahil (approved):

PR #2677, opened 2025-02-12
feat: Adapter for Desktop Background RPC service

nsrCodes:
Creates a base adapter class for an RPC like service created in the background. https://github.com/requestly/requestly-desktop-app/pull/129


nsrCodes:
Here's an example for how this can be used - https://github.com/requestly/requestly/pull/2678

PR #2687, opened 2025-02-13
[ENGG-2547] Improvement: error messaging in API client

nafees87n:
## Summary of changes
- Improved error state design
- Show warning for undefined variables if used in the request
- Propagate actual error from extension and [desktop app](https://github.com/requestly/requestly-desktop-app/pull/132)
- Pre validate cases where request could fail like no internet connection, invalid protocol, invalid method etc.
- Desktop app proxy is better at handling the network errors while extension propagates a general message like "Failed to fetch".

## Demo
https://github.com/user-attachments/assets/92e8440a-c8f9-4987-9922-866bd6c88868


rohanmathur91 (changes requested):
No error thrown for invalid URL and it still makes a request.
URL: `http//example.com` (any URL with missing `:` (colon) )

https://github.com/user-attachments/assets/03252fd8-7d3f-46b1-8429-9f30f1e4b645

rohanmathur91 (approved):

PR #2737, opened 2025-02-27
[ENGG-3115]: RBAC - sessions

rohanmathur91:
- Listing:
  - Remove all top-right primary actions.
  - Remove all share and delete options
- Saved view:
  - Make breadcrumb read only
  - Block all the editing options for all the inputs
- Draft view:
  - Block save button and on hover show view only message
  - Block all the inputs


app/src/features/sessionBook/screens/DraftSessionScreen/components/DraftSessionViewer/DraftSessionViewer.tsx
```diff
@@ -99,7 +103,12 @@ export const DraftSessionViewer: React.FC<DraftSessionViewerProps> = ({ isDeskto
               </RQButton>
             )}
 
-            <SaveSessionButton onSaveClick={handleSaveSessionClicked} />
+            <RQTooltip
```
iostreamer-X, line 106: This means a tooltip component always covers the button. Why? 
This wrapping itself should be conditional.

rohanmathur91: Tooltip itself is conditional, if we pass null to it, its not wrapped then.

iostreamer-X (approved):

PR #2759, opened 2025-03-05
[ENGG-3109] fix: unsaved tab changes showing up in case of existing requests

nafees87n:
Cause for unsaved changes triggerering:
1. Query params getting synced with URL after the creation of entry object.
2. Body being formatted on editor init.

Fixes
1. Query params are synced with URL at the time of entry object creation.
2. Body will not format on editor init. It is not needed as API client body already saves with formatted code


app/src/features/apiClient/screens/apiClient/components/clientView/APIClientView.tsx
```diff
@@ -140,21 +139,10 @@ const APIClientView: React.FC<Props> = ({ apiEntry, apiEntryDetails, notifyApiRe
...
-        request: {
-          ...apiEntry.request,
-          ...syncQueryParams(
-            apiEntry.request.queryParams,
-            apiEntry.request.url,
-            isEmpty(apiEntry.request.queryParams) ? QueryParamSyncType.TABLE : QueryParamSyncType.SYNC
-          ),
-        },
-      });
+    if (entry) {
```
RuntimeTerror10, line 142: Shouldn't this be `apiEntry`?

PR #2763, opened 2025-03-05
chore: add debug logs

nsrCodes:


nsrCodes:
was only made to debug in beta

PR #2791, opened 2025-03-12
[ENGG-3169]: Add move to collection support in local sync

RuntimeTerror10:


iostreamer-X (changes requested):
Looks good, just take a look at that code improvement.

iostreamer-X (approved):

PR #2799, opened 2025-03-13
[ENGG-3243] chore: teams to workspace slice migration

wrongsahil:
Aim: To reduce conflicts in https://github.com/requestly/requestly/pull/2335 branch


app/src/store/slices/workspaces/slice.ts
```diff
@@ -0,0 +1,96 @@
...
+      state.activeWorkspacesMembers = {
+        ...state.activeWorkspacesMembers,
+        ...action.payload,
+      };
+    },
+  },
+});
+const { actions, reducer: workspaceReducer } = slice;
+
+export const workspaceReducerWithLocal = getReducerWithLocalStorageSync(ReducerKeys.WORKSPACE, workspaceReducer, [
```
nsrCodes, line 90: suggestion: rename to something like `CachedWorkspaceReducer` instead? 
since we already use `local` to denote a workspace type

app/src/layouts/DashboardLayout/MenuHeader/WorkspaceSelector/index.jsx
```diff
@@ -71,13 +71,14 @@ const getWorkspaceIcon = (workspaceName) => {
 const WorkSpaceDropDown = ({ menu, hasNewInvites }) => {
   // Global State
   const user = useSelector(getUserAuthDetails);
-  const currentlyActiveWorkspace = useSelector(getCurrentlyActiveWorkspace);
-  const isWorkspaceMode = useSelector(getIsWorkspaceMode);
-  const isLocalWorkspace = useSelector(getIsWorkspaceLocal);
+  const activeWorkspaceId = useSelector(getActiveWorkspaceId);
+  const activeWorkspace = useSelector(getActiveWorkspace);
```
nsrCodes, line 75: could we skip creating `activeWorkspaceId` and instead use `activeWorkspace.id` ?

app/src/store/slices/workspaces/slice.ts
wrongsahil: We are using this in all reducers. Right now keeping it in same convention. We can rename it later on

app/src/layouts/DashboardLayout/MenuHeader/WorkspaceSelector/index.jsx
wrongsahil: Didn't want to look into small optimizations right now. So kept it same for now

nsrCodes (approved):

PR #2804, opened 2025-03-17
fix: clientStorageService init bug

wrongsahil:


app/src/services/clientStorageService/hooks/useClientStorageService.ts
```diff
@@ -1,14 +1,22 @@
-import { useEffect } from "react";
+import Logger from "lib/logger";
 import { useSelector } from "react-redux";
 import { getAppMode } from "store/selectors";
 import { initClientStorageService } from "..";
-import Logger from "lib/logger";
+import { useEffect } from "react";
+
+let firstInit = false;
```
nsrCodes: please change this to ref if possible. I'm afraid this might get cleaned up by the GC.

nsrCodes (approved):

app/src/services/clientStorageService/hooks/useClientStorageService.ts
wrongsahil: Done

PR #2824, opened 2025-03-26
[ENGG-3276]: update tab sources

rohanmathur91:


app/src/componentsV2/Tabs/routes/index.tsx
nafees87n, line 1: Can we name this `ApiClientTabRoutes`?

PR #2852, opened 2025-04-01
[ENGG-3297] [ENGG-3292] fix: URL syncing on tab changes

nafees87n:


app/src/componentsV2/Tabs/store/tabServiceStore.ts
```diff
@@ -229,11 +237,3 @@ export const useTabServiceWithSelector = <T>(selector: (state: TabServiceState)
...
   return useStore(store, useShallow(selector));
 };
-
-/**
- * Usage: const openTab = useTabServiceStore().use.openTab()
- */
-export const useTabServiceStore = () => {
-  const store = useContext(TabServiceStoreContext);
-  return store;
-};
```
nafees87n, line 239: Removing this because it was not being consumed as a hook. It was not reactive to the changes in state

PR #2862, opened 2025-04-03
chore: add target to all anchor tags

nsrCodes:
- else was a bug in case of desktop app as the link would then be opened in the current renederer (with no way of going back to the app)
- did not enforce for `mailto` links as they already open the native email client


Dinex-dev (approved):
LGTM!

PR #2889, opened 2025-04-09
[ENGG-3356] fix: local sync tabs regression issues

RuntimeTerror10:


app/src/features/apiClient/screens/apiClient/components/clientView/components/Collection/CollectionView.tsx
```diff
@@ -101,10 +110,19 @@ export const CollectionView: React.FC<CollectionViewProps> = ({ collectionId })
         }
 
         onSaveRecord(result.data, "open");
+        const wasForceRefreshed = await forceRefreshApiClientRecords();
+        if (wasForceRefreshed) {
+          closeTab(
+            new CollectionViewTabSource({
```
iostreamer-X, line 116: Let's see if we can remove this and move away from `closeTab`

PR #2924, opened 2025-04-22
[ENGG-3395] fix: shared list table row breaking

Aarushsr12:
- before
Screenshot 2025-05-19 at 4 26 53 PM

- after
image
image


rohanmathur91 (changes requested):
Rule names are not vertically centered in the screenshot provided

rohanmathur91:
Check if other fields have too much vertical padding, if yes the adjust accordingly, try to match it with rules table.

rohanmathur91 (approved):

PR #2929, opened 2025-04-24
[ENGG-3389] fix: unsave changes on existing rule

Aarushsr12:


app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -65,6 +66,7 @@ const Editor: React.FC<EditorProps> = ({
   isFullScreen = false,
   onFullScreenChange = () => {},
   hideToolbar = false,
+  effectTriggered,
```
nafees87n, line 69: This shouldn't be passed as props but an intristic property of the editor.

nafees87n (changes requested):

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
Aarushsr12: Parent component of Editor holds the power to update the states & redux dispatches, all the unsave changes/updation are controlled in response body component 

1. we need a property that tells outside of editor that effect has ran so the unsave flag value changes, since its handler is outside the editor
2. If we want to make the effectTriggered inside editor, I think we need to move the dispatching action inside the editor component

Aarushsr12: also there is seperation in terms of taking the input value & what to update, unsave or not in two different components

nsrCodes:
Saw this while debugging a separate issue. 

IMO the right solution would be to add back the `triggerUnsavedChanges` argument [here](https://github.com/requestly/requestly/blob/master/app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx#L40) and use it properly in the Request/Response body consumer _(and in extension all other consumers of the editor component)_

This is still being passed [for the Script rule](https://github.com/requestly/requestly/blob/master/app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/CustomScriptRow/index.jsx#L225) but not in other editors so it might be possible that it was missed while rewriting/migrating the editor component

app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/ResponseBodyRow/index.js
```diff
@@ -201,7 +201,7 @@ const ResponseBodyRow = ({ rowIndex, pair, pairIndex, ruleDetails, isInputDisabl
           "response.type": pair.response.type,
           "response.value": responseBodyValues.current[pair.response.type],
         },
-        triggerUnsavedChangesIndication: !codeFormattedFlag.current,
+        triggerUnsavedChangesIndication: triggerUnsavedChanges ? !codeFormattedFlag.current : codeFormattedFlag.current,
       })
```
RuntimeTerror10, line 205: No need to use `codeFormattedFlag` now. Remove this flag from code and just use `triggerUnsavedChanges`

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -74,6 +74,7 @@ const Editor: React.FC<EditorProps> = ({
   const toastOverlay = useMemo(() => allEditorToast[scriptId], [allEditorToast, scriptId]); // todo: rename
   const [isCodePrettified, setIsCodePrettified] = useState(false);
   const isDefaultPrettificationDone = useRef(false);
+  const isUnsaveChange = useRef(true);
```
RuntimeTerror10, line 77: This should be `false`

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -156,6 +159,9 @@ const Editor: React.FC<EditorProps> = ({
         if (!isDefaultPrettificationDone.current && prettifyOnInit) {
           await applyPrettification();
           isDefaultPrettificationDone.current = true;
+        } else {
+          // Typing in editor, no prettification, mark as unsaved change
+          isUnsaveChange.current = true;
```
RuntimeTerror10, line 164: This effect is only for prettifying the code once editor is initialized. Also this effect assumes that any further changes is going to be an unsaved change, which is not always true. 

We can remove this else from here and add a `onKeyDown` listener to Editor component which will set this flag to `true`

RuntimeTerror10 (changes requested):

RuntimeTerror10:
Also update the function name for `updateContent`. This is quite confusing with the `handleChange` function

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -128,13 +128,19 @@ const Editor: React.FC<EditorProps> = ({
     }
   };
 
-  const updateContent = useCallback((code: string): void => {
+  /*
+  (fx) sets the implicit change in the editor, prettification change is implicit change
+  Typing edits in editor is controlled by handleChange
+  */
+  const setImplicitChange = useCallback((code: string): void => {
```
RuntimeTerror10, line 135: Rename this to `handleEditorSilentUpdate`

RuntimeTerror10 (approved):

PR #2968, opened 2025-05-07
fix: block updates to sample rule

nsrCodes:
- This PR tends to solve the issue of the sample script rule's data being corrupted by the `<script>` tag artifacts being added. 
- This is was a brittle system, that was made robust by adding the validation very closely to the record saving methods.
- The `updateRulePairAtGivenPath` action is being called even for sample rules. 
- Updates made via sample rules form deviations in the UX and the steps to be followed to save the rule.
- This needs to be solved wholistically later. But is leading to silent regressions (like the one stated above) and I guess also the issue [this PR](https://github.com/requestly/requestly/pull/2929) intends to fix as well

### DIFF
--- a/app/src/store/features/rulesActions.ts
+++ b/app/src/store/features/rulesActions.ts
@@ -1,6 +1,5 @@
 import { getFilterObjectPath } from "utils/rules/getFilterObjectPath";
-//@ts-ignore
-import { CONSTANTS as GLOBAL_CONSTANTS } from "@requestly/requestly-core";
+// import { CONSTANTS as GLOBAL_CONSTANTS } from "@requestly/requestly-core";
 import { get, set } from "lodash";
 import { GlobalSliceState } from "store/slices/global/types";
 import { Group, QueryParamRule, Rule, ScriptRule } from "@requestly/shared/types/entities/rules";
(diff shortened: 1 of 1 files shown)


app/src/store/features/rulesActions.ts
```diff
@@ -151,17 +150,28 @@ export const updateRulePairAtGivenPath = (
...
+        (prevState.rules.currentlySelectedRule.data as Rule)?.pairs?.[pairIndex],
+        getFilterObjectPath(modificationPath),
+        value
+      );
+      updated = true;
+    }
   }
 
-  if (triggerUnsavedChangesIndication) {
+  if (updated && triggerUnsavedChangesIndication) {
```
nafees87n, line 165: Instead of maintaining the `updated` flag, can't we directly use `isRuleModificationAllowed` ?

nsrCodes: Since this is a very core function I was trying to keep it flexible. 

updates do have multiple entries which, in the future, could only partially be allowed. 
Hence wrote it this way.

unsaved changes need to be triggered even when partial updates are allowed.

P.S. Just anticipating that there could be further checks/fixes here in the future because of the re-rendering issue. But yes, certainly could be made simpler right now. wdyt?

nafees87n: the `updated` flag is bit hard to read and understand. Its usage is not clear just by looking at the code. `isRuleModificationAllowed` conveys the usage better

nafees87n (approved):

PR #2998, opened 2025-05-15
fix: mock not found - assign MockId to data

Aarushsr12:


app/src/backend/mocks/getMock.ts
```diff
@@ -27,7 +27,7 @@ const getMockFromFirebase = async (mockId: string): Promise<RQMockSchema> => {
 
   if (snapshot.exists()) {
     const data = snapshot.data() as RQMockSchema;
-    return data;
+    return { ...data, id: mockId };
```
wrongsahil, line 30: use `snapshot.id`

wrongsahil, line 30: This need to be done while fetching the list of mocks too

wrongsahil (changes requested):

app/src/backend/mocks/getMock.ts
Aarushsr12: just little confused here, why snapshot.id ? we are already assign that to data so ` { ...data, id: snapshot.id };` seems wrong?

wrongsahil: This is to handle cases in which we aren't able to set id.

app/src/backend/mocks/getMocks.ts
```diff
@@ -41,7 +41,7 @@ const getMocksFromFirebase = async (ownerId: string, type?: MockType): Promise<R
     if (!data?.id) {
       erroredMockIds.push(doc.id);
     } else {
-      result.push(doc.data());
+      result.push({ ...data, id: doc.id });
     }
   });
```
wrongsahil, line 46: This won't let the else condition run when data?.id isn't present.
```
if(!data?.id) {
          erroredMockIds.push(doc.id);
}
result.push({...data, id: doc.id});
```

wrongsahil (changes requested):

wrongsahil (changes requested):

wrongsahil (approved):

PR #3013, opened 2025-05-19
[ENGG-3627] fix: inconsistent extension enabled state

nafees87n:
The extensionEnabled state is stored in a variable in extension storage. So, if the extension storage is cleared (happens on workspace switch and logout). The extensionEnabled state resets.
This is one of the causes for extension status inconsistency.

### Solution
Do not clear `isExtensionEnabled` variable from extension storage on storage clearance. It should reset only on extension reload/uninstallation

Also added support for realtime updatation of UI on extension status change

This introduced a circular dependency which was resolved with #3019


browser-extension/common/src/storage.ts
```diff
@@ -40,6 +40,14 @@ export const clearAllRecords = async (): Promise<void> => {
   await chrome.storage[STORAGE_TYPE].clear();
 };
 
+export const clearAllRecordsExceptVariables = async (): Promise<void> => {
+  const allRecords = await getSuperObject();
+  const variableKeys = Object.keys(allRecords).filter((key) => key.startsWith("rq_var_"));
+  const variableRecords = Object.fromEntries(variableKeys.map((key) => [key, allRecords[key]]));
+  await chrome.storage[STORAGE_TYPE].clear();
+  await saveObject(variableRecords);
```
wrongsahil, line 48: Can we do something like this (not exactly this)?

```suggestion
  const extensionStatus = getVariable(Variable.IS_EXTENSION_ENABLED);
  await chrome.storage[STORAGE_TYPE].clear();
  setVariable(Variable.IS_EXTENSION_ENABLED, extensionStatus);
```

nafees87n: What if in future we also want to preserve some other variable

wrongsahil: Let's backup only extension enabled for now. In case we require more, we can modify this to be more generalized later.

wrongsahil: Lets get and set this key only

browser-extension/common/src/storage.ts
```diff
@@ -40,6 +41,21 @@ export const clearAllRecords = async (): Promise<void> => {
...
+  const variableRecords: StoreObject = {};
+  VARIABLE_KEYS.forEach((key, idx) => {
+    if (typeof variableValues[idx] !== "undefined") {
+      variableRecords[key] = variableValues[idx];
+    }
+  });
+
+  await chrome.storage[STORAGE_TYPE].clear();
+  await saveObject(variableRecords);
+};
```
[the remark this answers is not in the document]
nafees87n: Okay

browser-extension/common/src/storage.ts
```diff
@@ -37,7 +37,13 @@ export const removeRecord = async (key: string): Promise<void> => {
 };
 
 export const clearAllRecords = async (): Promise<void> => {
+  const VAR = "rq_var_isExtensionEnabled";
+  const isExtensionEnabledVariableValue = await getRecord(VAR);
```
wrongsahil, line 41: Why not just use `setVariable` and `getVariable` methods? It contains the prefix too?

wrongsahil (approved):

browser-extension/mv3/src/content-scripts/app/messageHandler.ts
```diff
@@ -67,6 +68,17 @@ const sendResponse = (originalEventData: ContentScriptMessage, response?: unknow
   });
 };
 
+const clearAllRecordsExpectExtensionEnabledVariable = async () => {
+  const isExtensionEnabled = await getVariable<boolean>(Variable.IS_EXTENSION_ENABLED);
+
+  clearAllRecords();
```
wrongsahil, line 74: Shouldn't this be awaited?

wrongsahil (approved):

wrongsahil (approved):

PR #3015, opened 2025-05-20
[ENGG-3633] fix: implicit Content-Type header missing when importing from Postman

RuntimeTerror10:


app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -151,6 +151,17 @@ const processAuthorizationOptions = (item: PostmanAuth.Item | undefined, parentC
...
+  switch (bodyType) {
+    case "json":
+      return RequestContentType.JSON;
+    case "text":
+      return RequestContentType.RAW;
+    case "html":
+      return RequestContentType.HTML;
+  }
+};
+
```
nsrCodes, line 164: We should cover all possible content types here. 
this seems to be missing cases for `xml` and `javascript`. 

Will also need to consider how body from such imports will be visible in the code editor

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -178,13 +189,32 @@ const createApiRecord = (
...
+    if (!isContentTypeHeaderSet) {
+      headers.push({
+        id: headers.length,
+        key: "Content-Type",
+        value: contentType,
+        isEnabled: true,
+      });
+    }
+  };
+
```
nsrCodes, line 203: let's convert this into a pure function and move it out of this function's body as a separate utility

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -178,13 +189,32 @@ const createApiRecord = (
...
+    contentType = getContentTypeForRawBody(options?.raw.language);
+    if (raw.length) {
+      handleAddImplicitContentTypeHeader(contentType);
+    }
   } else if (mode === "formdata") {
     contentType = RequestContentType.FORM;
+    /* 
+    TODO: handle formdata content-type addition
+    Postman calculates this header when request is made
+    */
```
nsrCodes, line 217: does this need to be addressed in this PR or has a ticket been made for it elsewhere?

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -194,6 +224,9 @@ const createApiRecord = (
       })) || [];
   } else if (mode === "urlencoded") {
```
nsrCodes, line 225: the content type header for when `mode` is `graphql` are not added.
not sure if the body is being handled either, could you please check that as well.

app/src/features/apiClient/types.ts
```diff
@@ -22,6 +22,7 @@ export enum RequestContentType {
   RAW = "text/plain",
   JSON = "application/json",
   FORM = "application/x-www-form-urlencoded",
+  HTML = "text/html",
```
nsrCodes, line 25: you will need to handle this new possible value throughout the code. 

most relevant here is the [request body](https://github.com/requestly/requestly/blob/master/app/src/features/apiClient/screens/apiClient/components/clientView/components/request/RequestBody.tsx) view. For eg. [here](https://github.com/requestly/requestly/blob/d28c216125f4a1c2665decc6e2ca3789a3617ae3/app/src/features/apiClient/screens/apiClient/components/clientView/components/request/RequestBody.tsx#L77-L96)

but there seem to be other instances as well

nsrCodes (changes requested):
please make sure that all possible cases are handled

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
rohanmathur91: Request content-type can be anything ([refer](https://www.iana.org/assignments/media-types/media-types.xhtml)), we should cover the most used ones, WDYT @nsrCodes?

Also, since we support `Raw` format for request body, then sever should be able to handle by itself (confirm this), in this case we dont need to take any action on how it will be visible in code editor

rohanmathur91: Request content-type can be anything ([refer](https://www.iana.org/assignments/media-types/media-types.xhtml)), we should cover the most used ones, WDYT @nsrCodes?

Also, since we support `Raw` format for request body, then sever should be able to handle by itself (confirm this), in this case we dont need to take any action on how it will be visible in code editor

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
RuntimeTerror10: This PR only addresses the missing content-type header which Postman adds internally. There 2 more phases after this which have their tickets
* Content-Type header for `form data`
* Adding support for other languages like `HTML`, `XML`, `JavaScript` in request body editor. (Currently `Text` is default for all unsupported raw types)

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
RuntimeTerror10: We don't support `graphql` mode yet. `Raw` mode with `text` body will be visible for now

app/src/features/apiClient/types.ts
RuntimeTerror10: This requires design input, for now all other modes will be shown in raw `text` mode.

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
nsrCodes: @rohanmathur91  I mentioned these specifically from the list of options provided by postman.
CleanShot 2025-05-20 at 17 18 02@2x

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -151,6 +157,107 @@ const processAuthorizationOptions = (item: PostmanAuth.Item | undefined, parentC
...
+  switch (mode) {
+    case PostmanBodyMode.RAW:
+      return processRawRequestBody(raw, options, headers);
+    case PostmanBodyMode.FORMDATA:
+      return { ...processFormDataBody(formdata), headers };
+    case PostmanBodyMode.URL_ENCODED:
+      return processUrlEncodedBody(urlencoded, headers);
+    default:
+      return {
+        requestBody: null,
```
nsrCodes, line 254: > Raw mode with text body will be visible for now [_Ref_](https://github.com/requestly/requestly/pull/3015#discussion_r2097742166)

doesn't seem like it. Body just gets lost here (both in the data and in the UI). I tried with the current state of these changes and [this export file ](https://github.com/user-attachments/files/20347553/misc.postman_collection.json)

CleanShot 2025-05-20 at 18 00 31@2x

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -151,6 +157,107 @@ const processAuthorizationOptions = (item: PostmanAuth.Item | undefined, parentC
...
+const getContentTypeForRawBody = (bodyType: string) => {
+  switch (bodyType) {
+    case "json":
+      return RequestContentType.JSON;
+    case "text":
+      return RequestContentType.RAW;
+    default:
+      return RequestContentType.RAW;
+  }
+};
```
nsrCodes, line 169: isn't this meant to give the content-type for each preview mode that [postman gives](https://github.com/requestly/requestly/pull/3015#discussion_r2097751425)?

since [you did](https://github.com/requestly/requestly/pull/3015/files#diff-3ed7caa53e23f71cd53c5431d592bff2ee456fec0b022083462e59f3e1b5bbd6R26-R27) add all the possible values of the enum now, why aren't we handling all them here?

nsrCodes (changes requested):
please test changes after addressing comments and make sure:
1. parts of the import data are not lost
2. UI does not break in any case
3. goal of the fix is achieved for all possible cases

app/src/features/apiClient/types.ts
nsrCodes: please make sure there is a ticket for it and that it has a stakeholder assigned to it.

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
nsrCodes: I understand that you [mentioned](https://github.com/requestly/requestly/pull/3015#discussion_r2097710818) there is a ticket for this
> Adding support for other languages like HTML, XML, JavaScript in request body editor. (Currently Text is default for all unsupported raw types)

but it seems trivial to set the correct content type here, so why not do it?

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
rohanmathur91: Got it.

PR #3019, opened 2025-05-20
fix: [extension] circular dependency among context menu and utils

nafees87n:


browser-extension/mv3/src/service-worker/services/utils.ts
```diff
@@ -187,3 +196,22 @@ export const getAppTabs = async (): Promise<chrome.tabs.Tab[]> => {
 
   return appTabs;
 };
+
+export const sendMessageToApp = async (messageObject: unknown) => {
+  const appTabs = await getAppTabs();
+  return Promise.all(appTabs.map(({ id }) => chrome.tabs.sendMessage(id, messageObject)));
+};
```
wrongsahil, line 203: Do we need to move this here after moving `updateActivationStatus` ? Isn't `updateActivationStatus` enough to break the chain?

browser-extension/mv3/src/service-worker/services/contextMenu.ts
```diff
@@ -1,34 +1,7 @@
...
-    title: isExtensionEnabled ? ToggleActivationStatusLabel.DEACTIVATE : ToggleActivationStatusLabel.ACTIVATE,
-  });
-
-  if (isExtensionEnabled) {
-    extensionIconManager.markExtensionEnabled();
-  } else {
-    extensionIconManager.markExtensionDisabled();
-  }
-
-  sendMessageToApp({ action: CLIENT_MESSAGES.NOTIFY_EXTENSION_STATUS_UPDATED, isExtensionEnabled });
```
wrongsahil, line 30: Instead of moving all this, can moving out `sendMessageToApp` to `toggleExtensionStatus` fix this?

wrongsahil, line 30: Instead of moving all this, can moving out `sendMessageToApp` to `toggleExtensionStatus` fix this?

wrongsahil (changes requested):

browser-extension/mv3/src/service-worker/services/utils.ts
nafees87n: Yes. Its necessary otherwise `utils` <-> `messageHandler` dependency will be present because `updateActivationStatus` depends on `sendMessageToApp`

browser-extension/mv3/src/service-worker/services/contextMenu.ts
nafees87n: Nope. `utils` uses `updateActivationStatus` from `contextMenu` while `contextMenu` uses `sendMessageToApp` from `utils`

Thats why its important to move both of them

wrongsahil: But sendMessageToApp is in messageHandler

nafees87n: `utils` <-> `messageHandler` dependency will be present because `updateActivationStatus` depends on `sendMessageToApp`

nafees87n: Actually utils shouldn't be importing from `contextMenu` or `messageHandler`. Thats why its needed to move both

wrongsahil (approved):

PR #3034, opened 2025-05-22
ENGG-3656: Fix error messaging in curl importer modal

RuntimeTerror10:
Screenshot 2025-05-22 at 6 09 07 PM


app/src/features/apiClient/screens/apiClient/components/modals/ImportFromCurlModal/ImportRequestModal.tsx
```diff
@@ -57,9 +70,14 @@ export const ImportRequestModal: React.FC<Props> = ({ isOpen, handleImportReques
       onOk={onImportClicked}
       maskClosable={false}
       confirmLoading={isRequestLoading}
-      onCancel={onClose}
+      onCancel={handleClose}
```
nsrCodes, line 73: I think the `error` also needs to be cleared in other cases as well

it is leading to intermittent behaviors like these 

https://github.com/user-attachments/assets/c79c6b5c-8435-42c4-af06-861166c37556

nsrCodes (approved):

PR #3040, opened 2025-05-22
Added support for global rejection handlers

iostreamer-X:


iostreamer-X:
To circumvent sentry aggressively taking over [onunhandledrejection](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event), we setup our own handler before sentry gets to do it.

This seems to affect the order in which handlers would be invoked quite severely. One would expect that once an event like "unhandledrejection" is emitted for a rejected promise, all the handlers would be invoked in one go.

But that's not the case in browsers, registration order seems to affect the priority and the time difference could vary from 1-3ms.

So to do anything, you have to be the first. And that's what this PR enables.

The handler that we registered manually is the first one and hence would always get priority. In this handler we check if any one has registered some code to be executed on an unhandled promise rejection, and if so, we execute it.

Our error boundary registers the code where the error is decorated with the `tags` property, expecting that before we send error to sentry, we will extract this property and tell sentry about it.

So timeline becomes:
- `t0`: Our unhandled promise rejection handler registered
- `t1`: Sentry's unhandled promise rejection  handler registered
- `t3`: An async error occurs
- `t4`: Our handler gets invoked and adds `tags` property to error
- `t6`: Sentry's handler gets invoked, and before sending the error, it handles our error's `tag` property

This setup allows us to make sure that the error will always have `tags` property before sentry attempts to send it.

PR #3074, opened 2025-05-29
draft: successful poc of autocompletes from a namespace file

nsrCodes:
requires adding `import {rq} from "./rq-namespace"` right now in the editor, but the setup works!

slightly buggy demo

https://github.com/user-attachments/assets/6f264c84-c6d4-4ac0-ad05-6622a04d6976


nsrCodes:
demo of poc after fixes


https://github.com/user-attachments/assets/9d57dddf-feff-4f25-ac34-1a2b9caac982

PR #3101, opened 2025-06-05
[ENGG-3668] fix: using withCredentials property in modify response rule with XHR

### LINKED ISSUE #2936: bug: Modify API Response Rule Strips Cookies from Requests (Causing 401 Errors)
### Describe your issue?

When using the Modify API Response rule in the Requestly Chrome Extension, cookies are not being sent with the request, resulting in a `401 Unauthorized` error. This behavior occurs even though no CORS issues or redirect rules are present.

Additional Observations by the user:

* The problem seems to originate from the `ajaxRequestInterceptor.ps.js` script.
* Specifically, the `this.rqProxyXhr` object does not include cookies.
* The `withCredentials `property is set to false, preventing cookies from being attached to the proxied request.

### Repro steps

The user was using the modify API rule, using a fixed endpoint filtered for the POST method. The action is to replace the response with static json.
Cookie is present when the query runs without the rule enabled.
When the rule gets enabled, user gets the 401 - cookie is missing in the request

### What Requestly tool were you using?

Extension

- [ ] Desktop-App
- [ ] Android SDK
- [ ] Selenium

### Your Environment

MacOS

### Requestly Version

24.11.25

### Error screenshot

*No response*

nafees87n:
FIxes #2936 

Testing: https://github.com/nafees87n/poc/tree/master/cors-credentials-demo


wrongsahil:
@nafees87n Can you share some curls which can be used for testing this

wrongsahil (approved):

PR #3112, opened 2025-06-09
patch: react-codemirror

nsrCodes:
shared further details [here](https://github.com/uiwjs/react-codemirror/issues/694#issuecomment-2954215912)

### DIFF
--- a/app/patches/@uiw+react-codemirror+4.22.1.patch
+++ b/app/patches/@uiw+react-codemirror+4.22.1.patch
@@ -1,0 +1,162 @@
+diff --git a/node_modules/@uiw/react-codemirror/cjs/useCodeMirror.js b/node_modules/@uiw/react-codemirror/cjs/useCodeMirror.js
+index b63f975..8fb284f 100644
+--- a/node_modules/@uiw/react-codemirror/cjs/useCodeMirror.js
++++ b/node_modules/@uiw/react-codemirror/cjs/useCodeMirror.js
+@@ -157,6 +157,7 @@ function useCodeMirror(props) {
+     }
+     var currentValue = view ? view.state.doc.toString() : '';
+     if (view && value !== currentValue) {
++      console.log("cjs/ value chnged from ".concat(currentValue, " to ").concat(value));
+       view.dispatch({
+         changes: {
+           from: 0,
+diff --git a/node_modules/@uiw/react-codemirror/esm/useCodeMirror.js b/node_modules/@uiw/react-codemirror/esm/useCodeMirror.js
+index eab503a..4fe0770 100644
+--- a/node_modules/@uiw/react-codemirror/esm/useCodeMirror.js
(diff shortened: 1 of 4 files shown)


app/patches/@uiw+react-codemirror+4.22.1.patch
```diff
@@ -0,0 +1,75 @@
...
++      while (
++        endCurrent > start &&
++        endValue > start &&
++        value[endValue - 1] === currentValue[endCurrent - 1]
++      ) {
++        endCurrent--;
++        endValue--;
++      } 
++      const changesToDispatch = {
++        from: start,
```
iostreamer-X: We have now essentially moved start from its default value of `0` to something closer to the diff. Does this mean that the issue is still happening but is localized around the new start? Which in turn is closer to the change and hence not noticeable?

If there was to a synthetic test such that start always ends up being 0, would this issue resurface?

nsrCodes: yes it would because the dispatch does not give a selection here.

But for context this is only triggered when the value strays away from the internal code mirror state - [ref](https://github.com/requestly/requestly/pull/3112/files#diff-836582cf0985e45a84b277243e8f9cdc66e6a2fda86ba758a16be0fe1f58befaR24)

which would only happen if the synthetic test were to emulate a case where the content of the editor is changed in some way other than typing into the editor 

for e.g. a button that (for the sake of argument) partially clears the editor. 

that would surely lead to errors if triggered while user is still typing. But in that case it is expected that focus is first removed from the editor. 

if we do have a case where the `value` state is changed drastically, via somehwere else in the app, we have other things to account for that require taking a lot more control of codemirror inside our code (the first of which would be specifying where the cursor would end up)

iostreamer-X: Got it, looks good, let’s deploy on beta and test.

iostreamer-X: Btw, would the way we do prettification, count as a case where the content of the editor is changed in some way other than typing into the editor?

nsrCodes: yes it does, but like I mentioned, it requires the focus being removed from the editor.

Ideally the curor should end up after the word/symbol that it was next to before the prettification, but building that requires:
1. the prettification to be done via a view dispatch (and regardless of this case, the prettification should be done through a code mirror extension, it might be causing other regressions that we haven't faced yet)
2. us keeping track of and finding that position before and after the prettification 

we can triage if the ROI is worth putting in this effort for this case. 
But for now, yes prettification feels like the case mentioned in the example.

nsrCodes: > Got it, looks good, let’s deploy on beta and test.

deployed, seems to be working well so far

iostreamer-X (approved):

PR #3121, opened 2025-06-09
[ENGG-3736] chore: support block screen

wrongsahil:


app/src/componentsV2/BlockScreen/hooks/useIsUserBlocked.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  useEffect(() => {
+    if (!isLoggedIn) {
+      return;
+    }
+
+    const emailDomain = email?.split("@")[1];
+    const emailDomainKey = emailDomain?.replaceAll(".", "_dot_");
+    const rdb = getDatabase();
+    const domainBlockConfigRef = ref(rdb, `globalBlockConfig/domain/${emailDomainKey}`);
```
rohanmathur91, line 57: Are we going to handle this domains manually? what if we get a set of domains to be block, do we have any script which write here?

rohanmathur91 (approved):
LGTM!

app/src/componentsV2/BlockScreen/hooks/useIsUserBlocked.ts
wrongsahil: Manually for now.

PR #3125, opened 2025-06-10
[ENGG-3742] fix: [extension] include credentials in API requests 

nafees87n:
1. Support for include credentials to be configurable from UI
2. Default value is true (always include)
3. Have a fallback UI controlled by feature flag


app/src/features/apiClient/screens/apiClient/components/clientView/components/request/components/RequestTabs/RequestTabs.tsx
```diff
@@ -155,6 +156,24 @@ const RequestTabs: React.FC<Props> = ({
       items={tabItems}
       size="small"
       moreIcon={null}
+      tabBarExtraContent={
+        showCredentialsCheckbox ? (
```
iostreamer-X: Use `src/components/common/Conditional/index.tsx`

iostreamer-X (approved):

PR #3152, opened 2025-06-16
Make parseAuth common and use across

iostreamer-X:


app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -88,10 +88,10 @@ export const createAutogeneratedStore = (entry?: RQAPI.Entry) => {
     headerVersion: 0,
     queryParamsVersion: 0,
 
-    initialize: <T>(entry: T, resolver: <U extends Record<string, any>>(input: U) => T) => {
+    initialize: <T extends RQAPI.Entry>(entry: T, resolver: <U extends Record<string, any>>(input: U) => T) => {
```
Aarushsr12: Extend T to RQAPI.Entry and returning the same Entry, makes it bit tightly coupled? 

```
  const resolver = useCallback(
    <T extends Record<string, any>>(template: T) => {
      return renderVariables(template, apiEntryDetails?.collectionId).result;
    },
    [renderVariables, apiEntryDetails?.collectionId]
  );

  useEffect(() => {
    // On mounting only, we are intializing, so dep array is empty
    autoGeneratedStore.getState().initialize(entry, resolver);
  }, []);
```

resolver takes a Record object & returns a record object, without it getting typeError at line `278`

Aarushsr12: `initialize: <T>(entry: T, resolver: <U extends Record<string, any>>(input: U) => T)`
Making it bit more acceptable like this ?

Aarushsr12 (changes requested):

app/src/features/apiClient/store/autogenerateStore.ts
Aarushsr12: Instead of this, we can do this
```
    initialize: <T extends RQAPI.Entry>(entry: T, resolver: <U extends Record<string, any>>(input: U) => U) => {
      // it accepts an entry & returns an entry -> so types would be like that
      const result = resolver(entry);
      const namespaces = parseRequestEntry(result);
      set({
        namespaces,
      });
    },
```
Making resolver function to return U which is holding Record<string, any>, it solves typeError

iostreamer-X: Looks good

PR #3156, opened 2025-06-17
[ENGG-3788] show all response headers for API client requests

nafees87n:
- All the response headers were not visible for API client requests. Headers like `set-cookie` `x-frame-options` were not visible
- Introduced a mechanism to relay response headers from webRequest interceptor to the API client
- FYI: It only works in chromium based browsers and not firefox. Firefox does not allow [`extraHeaders` options](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onHeadersReceived)


browser-extension/mv3/src/service-worker/services/apiClient/index.ts
```diff
@@ -128,6 +140,7 @@ export async function getAPIResponse(apiRequest: Request): Promise<Response | {
       redirectedUrl: response.url !== url ? response.url : "",
     };
   } catch (e) {
+    apiRequestCorrelationManager.removeHandler(requestlyId);
```
iostreamer-X: Should be in `finally`?

browser-extension/mv3/resources/static-rules/headerRules.json
```diff
@@ -0,0 +1,17 @@
...
+  {
+    "id": 1,
+    "condition": {
+      "resourceTypes": ["xmlhttprequest"]
+    },
+    "action": {
+      "type": "modifyHeaders",
+      "requestHeaders": [
+        {
+          "header": "x-requestly__id__",
```
wrongsahil: Can we call this `x-requestly-id`

browser-extension/mv3/src/service-worker/services/apiClient/index.ts
```diff
@@ -89,6 +96,16 @@ export async function getAPIResponse(apiRequest: Request): Promise<Response | {
...
     const requestStartTime = performance.now();
+    let responseHeaders: KeyValuePair[] = [];
+    apiRequestCorrelationManager.addHandler(
+      requestlyId,
+      (requestDetails: chrome.webRequest.WebResponseHeadersDetails) => {
+        responseHeaders = requestDetails.responseHeaders.map((header) => ({
+          key: header.name,
+          value: header.value,
+        }));
+      }
```
wrongsahil: As this is async, is it possible that these are executed after the response is returned to the UI?

browser-extension/mv3/src/service-worker/services/apiClient/ApiRequestCorrelationManager.ts
```diff
@@ -0,0 +1,44 @@
+class ApiRequestCorrelationManager {
+  private handlers: Map<string, Function> = new Map();
+  private registeredHandlers: Map<string, Function> = new Map();
```
wrongsahil: [Nitpick] We can keep something like this too. Easier to understand imo


```
private handlers: Map<string, Function> = new Map(); // rqId: handler
private webRequestIdToRQId: Map<string, string>; // This establishes the correlation between our RQ-id and webRequestId

```

wrongsahil (changes requested):

browser-extension/mv3/src/service-worker/services/apiClient/index.ts
nafees87n: `fetch` awaits for the response being returned. We listen for the response headers on `onHeadersReceived` which is triggered before any response is received. 

In [webRequest terminology](https://developer.chrome.com/docs/extensions/reference/api/webRequest#life_cycle_of_requests), we listen on `webRequest.onHeadersReceived` and fetch is resolved when `webRequest.onCompleted` is triggered. This ensures that response headers are always available before the response is received.

We can also read response headers from fetch response as a fallback.

wrongsahil: Can you try adding a delay in headersReceived and see if what happens? 

AFAIK, now we are using non blocking APIs so this would return nothing

nafees87n: Delay wont work. Its async.
We are just using the sequence of events because `fetch` call can be awaited

browser-extension/mv3/src/service-worker/services/apiClient/index.ts
```diff
@@ -89,6 +96,18 @@ export async function getAPIResponse(apiRequest: Request): Promise<Response | {
 
   try {
     const requestStartTime = performance.now();
+
+    let responseHeaders: KeyValuePair[] = [];
+    apiRequestCorrelationManager.addHandler(
+      requestlyId,
+      (requestDetails: chrome.webRequest.WebResponseHeadersDetails) => {
+        responseHeaders = requestDetails.responseHeaders.map((header) => ({
```
wrongsahil: `requestDetails.responseHeaders?.map` add undefined check.

According to the type, it can be undefined

wrongsahil (approved):

PR #3160, opened 2025-06-18
[ENGG-3727] fix(style): update font size as per design system

rohanmathur91:


browser-extension/common/src/popup/components/DesktopAppProxy/components/ConnectedToDesktopView/connectedToDesktopView.scss
```diff
@@ -19,6 +19,6 @@
   .connected-description {
     color: var(--alias-tokens-color-text-subtle, #bbb);
     text-align: center;
-    font-size: var(--font-size-sm, 13px);
+    font-size: var(--requestly-font-size-sm, 13px);
```
wrongsahil, line 22: This will break as there is no design system variable in extension

PR #3126, opened 2025-06-10
feat: add os to the compatibility checks

nsrCodes:

PR #3178, opened 2025-06-23
chore: update to have one screen for both incompatible and breaking state

nsrCodes:
~~will need to be rebased when https://github.com/requestly/requestly/pull/3126 is merged~~

dialog demo (github asset upload keeps failing for some reason):
https://youtu.be/oe24Fj4HCfU


app/src/components/mode-specific/desktop/UpdateDialog/MandatoryUpdateScreen.scss
```diff
@@ -0,0 +1,28 @@
+.full-screen-dialog {
```
wrongsahil, line 1: CleanShot 2025-06-24 at 14 04 58@2x

Top padding looks off. We should add some padding on top

app/src/components/mode-specific/desktop/UpdateDialog/index.js
```diff
@@ -51,48 +49,20 @@ const UpdateDialog = () => {
...
-      );
-    } else if (isUpdateDownloaded) {
-      return <NonBlockingDialog updateDetails={updateDetailsRef.current} quitAndInstall={quitAndInstall} />;
-    }
-  }
-  return null;
+  if (appMode !== GLOBAL_CONSTANTS.APP_MODES.DESKTOP || !isUpdateAvailable) return null;
+
+  if (isBreaking()) return <MandatoryUpdateScreen handleCTAClick={redirectToDownloadPage} CTAText="Download Now" />;
+  if (isIncompatible()) return <MandatoryUpdateScreen handleCTAClick={quitAndInstall} CTAText="Install Now" />;
```
wrongsahil: This doesn't handle the background download progress of the auto update. Won't that be a problem here?

app/src/components/mode-specific/desktop/UpdateDialog/index.js
```diff
@@ -51,48 +49,20 @@ const UpdateDialog = () => {
...
-          isUpdateDownloaded={isUpdateDownloaded}
-          updateProgress={updateProgress}
-        />
-      );
-    } else if (isUpdateDownloaded) {
-      return <NonBlockingDialog updateDetails={updateDetailsRef.current} quitAndInstall={quitAndInstall} />;
-    }
-  }
-  return null;
+  if (appMode !== GLOBAL_CONSTANTS.APP_MODES.DESKTOP || !isUpdateAvailable) return null;
```
wrongsahil, line 64: `!isUpdateAvailable` won't be true in case of windows desktop app since we aren't supporting auto updates. Won't this cause issues?

nsrCodes: The update will be available in that case. it's just that it won't be installable. 

the `isUpdateAvailable` is set based on if there is a latest release after the current one for that platform.

app/src/components/mode-specific/desktop/UpdateDialog/MandatoryUpdateScreen.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <img src="/assets/media/components/force-update.svg" alt="Update now" />
+          </Col>
+          <Col span={19} className="info">
+            <Title level={5}>Update required: This version is no longer supported</Title>
+            <Row>
+              <Text type="secondary">
+                You're using an outdated version of Requestly that is no longer functional. To continue using the app,
+                please download and install the latest version below.
+              </Text>
+            </Row>
```
wrongsahil: This description might misguide user when only the app needs to be restarted.

app/src/components/mode-specific/desktop/UpdateDialog/index.js
```diff
@@ -51,48 +49,21 @@ const UpdateDialog = () => {
...
+
+  if (isBreaking()) return <MandatoryUpdateScreen handleCTAClick={redirectToDownloadPage} CTAText="Download Now" />;
+  if (!isUpdateDownloaded) return null;
```
wrongsahil, line 68: Until the update is downloaded in the background, user won't see the screen. 
Isn't the aim of the screen is to block user from doing anything? If yes, Should we bring back the block dialog with the download progress?

wrongsahil (approved):

PR #3191, opened 2025-06-25
[ENGG-3869] fix: update regex to support env. variables

Aarushsr12:


app/src/features/apiClient/constants.ts
```diff
@@ -68,4 +68,4 @@ export const DEFAULT_SCRIPT_VALUES = {
     "// **********************************************\n// 🛠️ Use JavaScript to visualize responses: https://docs.requestly.com/general/api-client/scripts\n// **********************************************\n",
 };
 
-export const INVALID_KEY_CHARACTERS = /[^!#$%&'*+\-.^_`|~0-9A-Za-z]/;
+export const INVALID_KEY_CHARACTERS = /^(?!\{\{.*\}\}|^\{\{?$|^\}\}?$)[^!#$%&'*+\-.^_`|~0-9A-Za-z]/;
```
nafees87n, line 75: The regex is too complex. We can simplify this.
Now this is being done after parsing the request, so braces shouldn't be present here because template's values would have been consumed here. Right?

app/src/features/apiClient/constants.ts
nafees87n, line 1: According to the [specification](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6), only these chars 
```
"!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA 
```
are allow in header names.

We can specify a simply regex to check if the header name is composed only of these characters or not. 
We can use this regex `/^[!#$%&'*+\-.0-9A-Z^_`a-z|~]+$/`.

A similar check has been implemented in [here](https://github.com/requestly/requestly/blob/2ee18ad9d55d25adbb984ec6f43cefbbb7667bb0/app/src/views/features/rules/RuleEditor/components/Header/ActionButtons/CreateRuleButton/actions/index.js#L250-L306) as well. We should create a utility function to be used in both the checks.
Iff it increases the scope, we can skip changing the rules validation for now.

nafees87n (changes requested):

app/src/features/apiClient/helpers/apiClientExecutor/apiClientExecutor.ts
```diff
@@ -120,6 +119,11 @@ export class ApiClientExecutor {
   }
 
   private preValidateRequest() {
+    const INVALID_CHARACTERS = /[^!#$%&'*+\-.0-9A-Z^_a-z|~]/;
```
nafees87n:
```suggestion
    const INVALID_HEADER_CHARACTERS = /[^!#$%&'*+\-.0-9A-Z^_a-z|~]/;
```

WDYT?

app/src/features/apiClient/constants.ts
nafees87n, line 75: Add a comment that it looks for brace enclosures {{..}}

app/src/features/apiClient/helpers/apiClientExecutor/apiClientExecutor.ts
nafees87n: Add a comment to indicate which cases will fail with the `INVALID_KEY_CHARACTERS` constant here

Aarushsr12: Yes we can go for this too!

nafees87n (approved):

PR #3194, opened 2025-06-26
[ENGG-3864] fix: autogenerated headers table layout

Aarushsr12:


Aarushsr12:
#3204 changes moved here

PR #3199, opened 2025-06-26
[ENGG-3896]: syncing layer for api client logged out state 

rohanmathur91:


app/src/features/apiClient/helpers/modules/sync/localStore/services/LocalStoreRecordsSync.ts
```diff
@@ -20,7 +20,7 @@ export class LocalStoreRecordsSync implements ApiClientRecordsInterface<ApiClien
   }
 
   private getNewId() {
-    return uuidv4();
+    return generateDocumentId("apis");
```
iostreamer-X: Why not uuid?

app/src/features/apiClient/helpers/modules/sync/localStore/store/hooks.ts
```diff
@@ -0,0 +1,15 @@
+import { useStore } from "zustand";
+import { syncServiceStore } from "./useSyncServiceStore";
+import { APIClientSyncService } from "./types";
+import { useShallow } from "zustand/shallow";
+
+export const useSyncServiceStore = <T>(selector: (state: APIClientSyncService.State) => T) => {
```
iostreamer-X: Let's keep this consistent and call it `useSyncService`

app/src/features/apiClient/helpers/modules/sync/localStore/store/useSyncServiceStore.ts
```diff
@@ -0,0 +1,88 @@
+import { create } from "zustand";
+import { persist } from "zustand/middleware";
+import localStoreRepository from "../ApiClientLocalStorageRepository";
+import * as Sentry from "@sentry/react";
+import { APIClientSyncService } from "./types";
+
+export const createSyncServiceStore = () => {
+  const syncServiceStore = create<APIClientSyncService.State>()(
+    persist(
```
iostreamer-X: The initial state shouldn't come from local storage, nor should it be relied upon. Use the other index-db as the only source of truth.

app/src/features/apiClient/screens/apiClient/components/sidebar/APIClientSidebar.tsx
```diff
@@ -45,6 +46,8 @@ const APIClientSidebar: React.FC<Props> = () => {
     apiClientRecordsRepository,
   } = useApiClientContext();
 
+  useAutoSync();
```
iostreamer-X: If this is not using anything from this component, let's place it somewhere better than sidebar?

app/src/features/apiClient/helpers/modules/sync/useAutoSync.ts
```diff
@@ -0,0 +1,30 @@
...
+  const isSynced = useIsAllSynced();
+  const [syncAll, resetSyncStatus] = useSyncServiceStore((state) => [state.syncAll, state.resetSyncStatus]);
+
+  useEffect(() => {
+    if (!uid) {
+      resetSyncStatus();
+      return;
+    }
+
+    if (activeWorkspace?.id || isSynced) {
```
iostreamer-X: Let's add an explicit check for private workspaces here. Else we run the risk of syncing this into other workspaces.

iostreamer-X (changes requested):

app/src/features/apiClient/helpers/modules/sync/localStore/services/LocalStoreRecordsSync.ts
rohanmathur91: if in local mode we have uuid as id, while creating new api records we use existing ids only, at this stage we will have two different types of id formats, so to avoid that keeping it as firebase id only.

app/src/features/apiClient/helpers/modules/sync/localStore/store/useSyncServiceStore.ts
rohanmathur91: We await while clearing indexDb, so it make sure that the entity sync status is only updated once the indexDb is cleared.

app/src/features/apiClient/screens/apiClient/components/sidebar/APIClientSidebar.tsx
rohanmathur91: Apart from sidebar we have the `<ApiRecordsProvider/>`, but placing `useAutoSync` there will trigger both fetching and local record sync useEffects together, and we need a way to ensure that synced records are again fetched.

app/src/features/apiClient/helpers/modules/sync/localStore/components/AutoSyncLocalStoreDaemon.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const activeWorkspace = useSelector(getActiveWorkspace);
+  const uid = user?.details?.profile?.uid;
+  const syncRepository = useGetApiClientSyncRepo();
+  const [syncAll, resetSyncStatus] = useSyncService((state) => [state.syncAll, state.resetSyncStatus]);
+  const [addNewRecords] = useAPIRecords((state) => [state.addNewRecords]);
+  const [syncedLocalApiRecords, setSyncedLocalApiRecords] = useState<RQAPI.Record[]>([]);
+
+  useEffect(() => {
+    if (!uid) {
+      resetSyncStatus();
```
iostreamer-X: This sets sync as idle but doesn't actually cancel it right?

app/src/features/apiClient/store/apiRecords/apiRecords.store.ts
```diff
@@ -203,6 +204,12 @@ export const createApiRecordsStore = (initialRecords: { records: RQAPI.Record[];
       get().getRecordStore(record.id).getState().updateRecordState(record);
     },
 
+    addNewRecords(records) {
+      const updatedRecords = [...get().apiClientRecords, ...records];
+      get().refresh(updatedRecords);
+      get().updateRecords(updatedRecords);
```
iostreamer-X: This is excessive

PR #3204, opened 2025-06-27
[ENGG-3864] fix: Autogenerated header table layout 

Aarushsr12:


Aarushsr12:
> Can we avoid using important. Lets add important only when styles are added directly to tags else we can increase the specificity.

removed important from the css, they were unnecessarily added during dev and was not cleaned up, @RuntimeTerror10 please re-review now!

RuntimeTerror10 (approved):

PR #3205, opened 2025-06-29
[ENGG-3867] fix: Response status color

Aarushsr12:


app/src/lib/design-system/custom/base.css
```diff
@@ -253,6 +253,12 @@
 .warning {
   color: var(--warning);
 }
+.error {
+  color: var(--error);
+}
+.info {
+  color: var(--info);
+}
```
rohanmathur91: Don't consume this classes as they are from old design system, [refer](https://github.com/requestly/requestly/blob/master/app/src/lib/design-system-v2/helpers/theme.css) for updated variables.

rohanmathur91 (changes requested):

rohanmathur91 (approved):

PR #3207, opened 2025-06-30
fix: table cell color change on dropdown

Aarushsr12:


Aarushsr12:
https://stackoverflow.com/questions/61083813/how-to-avoid-internal-autofill-selected-style-to-be-applied

rohanmathur91:
Can we also fix this for sign in form? currently there is no autocomplete for email.

PR #3223, opened 2025-07-02
chore: submit deviceId attr

nsrCodes:


app/src/hooks/LocalUserAttributesHelperComponent.js
```diff
@@ -17,6 +19,8 @@ const LocalUserAttributesHelperComponent = () => {
     if (!userAttributes?.deviceId) {
       const deviceId = uuidv4();
       stableDispatch(globalActions.updateUserAttributes({ deviceId }));
```
wrongsahil, line 21: Should we directly fire the `submitAttrUtil(APP_CONSTANTS.GA_EVENTS.ATTR.CUSTOM_DEVICE_ID, userAttributes?.deviceId);` here only as local.ts does the same thing

wrongsahil (changes requested):

app/src/hooks/LocalUserAttributesHelperComponent.js
nsrCodes: @wrongsahil  I believe that would work if we were just rolling this out. But now, moving this here would miss the cases for sessions that already have a deviceId assigned to it. wdyt

PR #3249, opened 2025-07-07
[ENGG-3934]: add api client example collections import nudge

rohanmathur91:
https://github.com/user-attachments/assets/47fbee82-981c-45da-882d-f47bb7b8e56a

Screenshot 2025-07-08 at 1 02 24 PM


app/src/features/apiClient/screens/apiClient/components/sidebar/components/ExampleCollectionsNudge/ExampleCollectionsNudge.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export const ExampleCollectionsNudge: React.FC<ExampleCollectionsNudgeProps> = ({ fallback = null }) => {
+  const dispatch = useDispatch();
+  const user = useSelector(getUserAuthDetails);
+  const uid = user?.details?.profile?.uid;
+
+  const syncRepository = useApiClientRepository();
+  const { forceRefreshEnvironments } = useEnvironmentManager({ initFetchers: false });
+
+  const recordsStore = useAPIRecordsStore();
```
iostreamer-X, line 29: I think we should use useContext here to get the store from context, than this hook. Primarily because this is reactive, it will subscribe this component to the changes done in records store, which we are not interested in.

rohanmathur91: Its a new hook which just returns a vanilla store.

app/src/features/apiClient/screens/apiClient/components/sidebar/components/collectionsList/collectionRow/CollectionRow.tsx
```diff
@@ -272,11 +273,16 @@ export const CollectionRow: React.FC<Props> = ({
...
                     </div>
                   )}
-                  {isActive ? (
-                    <PiFolderOpen className="collection-expand-icon" />
-                  ) : (
-                    <MdOutlineFolder className="collection-expand-icon" />
-                  )}
+                  {
+                    // @ts-ignore
+                    record.isRoot ? (
```
iostreamer-X: Should be `record.isRoot && record.isExample`. 
I understand isRoot is only set for example, but isRoot is a very generic name, can be used by anyone else in future as well.

Let's keep it as `record.isExampleRoot`

PR #3252, opened 2025-07-08
fix: wrap appName filter generator to not crash for non parsable useragent

nsrCodes:
was leading to `URI malformed` errors. Can be reproduced by having this rule active inside desktop app.

[faulty UA header.json](https://github.com/user-attachments/files/21123783/faulty.UA.header.json)

This does show all such instances of user agent headers under one `unknown app` option in the sidebar filters as such:

CleanShot 2025-07-08 at 19 52 16@2x

Adding a manual exception to better understand what types of values are causing this crash

### DIFF
--- a/app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/index.jsx
+++ b/app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/index.jsx
@@ -49,6 +49,7 @@ import { doesStatusCodeMatchLabels, getGraphQLOperationValues } from "./utils";
 import { TRAFFIC_TABLE } from "modules/analytics/events/common/constants";
 import { trackRQDesktopLastActivity } from "utils/AnalyticsUtils";
 import { RQTooltip } from "lib/design-system-v2/components/RQTooltip/RQTooltip";
+import { captureException } from "backend/apiClient/utils";
 
 const CurrentTrafficTable = ({
   logs: propLogs = [],
@@ -489,7 +490,7 @@ const CurrentTrafficTable = ({
   const getLogAvatar = useCallback(
     (key, logName = "", avatarUrl) => {
       const isSelected = trafficTableFilters[key].includes(logName);
-
+      console.log("DG-2: all trafficTableFilters", trafficTableFilters);
       return (
         <RQTooltip mouseEnterDelay={0.3} placement="right" title={logName.length >= 20 ? logName : ""}>
           <Avatar size={18} src={avatarUrl} style={{ display: "inline-block", marginRight: "4px" }} />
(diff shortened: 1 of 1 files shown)


app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/index.jsx
```diff
@@ -489,7 +490,7 @@ const CurrentTrafficTable = ({
   const getLogAvatar = useCallback(
     (key, logName = "", avatarUrl) => {
       const isSelected = trafficTableFilters[key].includes(logName);
-
+      console.log("DG-2: all trafficTableFilters", trafficTableFilters);
```
wrongsahil: Remove console log

app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/index.jsx
```diff
@@ -513,10 +514,18 @@ const CurrentTrafficTable = ({
...
+        const avatarDomain = APPNAMES[logNameURI.split(" ")[0].toLowerCase()];
+        const avatarUrl = `https://www.google.com/s2/favicons?domain=${avatarDomain}`;
+        return getLogAvatar(key, logNameURI, avatarUrl);
+      } catch (e) {
+        console.log("faulty logname: ", logName);
+        captureException(e);
+        const avatarUrl = "https://www.google.com/s2/favicons?domain=randooooom.com"; // random url to show the unfound icon
+        const name = `unknown-${key}`;
+        return getLogAvatar(key, name, avatarUrl);
+      }
```
wrongsahil: - Uses sentry.captureMessage instead of exception
- Moved to finally instead of exception
```suggestion
      try {
          const logNameURI = decodeURIComponent(logName.trim());
       } catch (e) {
           Logger.log("faulty logname: ", logName);
           Sentry.captureMessage(e);
      } finally {
          const avatarDomain = APPNAMES[logNameURI.split(" ")[0].toLowerCase()];
          const avatarUrl = `https://www.google.com/s2/favicons?domain=${avatarDomain}`;
          return getLogAvatar(key, logNameURI, avatarUrl);
      }
```

wrongsahil (changes requested):

app/src/components/mode-specific/desktop/InterceptTraffic/WebTraffic/TrafficTableV2/index.jsx
nsrCodes: added a [slightly modified version](https://github.com/requestly/requestly/pull/3252/commits/107c8f25f8f3e828a39be6f5329d6186a8964a15) of this. 
Adding a `return` inside the `finally` was deemed unsafe by the linter 
Reference - https://eslint.org/docs/latest/rules/no-unsafe-finally

PR #3255, opened 2025-07-10
fix: sync fire and forget could still lead to multiple syncs

iostreamer-X:
https://github.com/user-attachments/assets/f895930d-17a0-473f-a10d-62eb01881792

### DIFF
--- a/app/src/features/apiClient/helpers/modules/sync/localStore/components/AutoSyncLocalStoreDaemon.tsx
+++ b/app/src/features/apiClient/helpers/modules/sync/localStore/components/AutoSyncLocalStoreDaemon.tsx
@@ -12,44 +12,23 @@ import {
   trackLocalStorageSyncFailed,
 } from "modules/analytics/events/features/apiClient";
 import * as Sentry from "@sentry/react";
-import { syncServiceStore } from "../store/syncServiceStore";
-import { APIClientSyncService } from "../store/types";
 
-export const AutoSyncLocalStoreDaemon: React.FC<{}> = () => {
-  const user = useSelector(getUserAuthDetails);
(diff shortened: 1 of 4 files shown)


app/src/features/apiClient/helpers/modules/sync/localStore/components/AutoSyncLocalStoreDaemon.tsx
```diff
@@ -82,7 +68,50 @@ export const AutoSyncLocalStoreDaemon: React.FC<{}> = () => {
...
+        await syncTask;
+        const records = getAllRecords();
+
+        // If there is no data, then there' no need to reset anything
+        if (!records.length) {
+          return;
+        }
+
+        // But if there are records, we reset them since syncing has been successful
+        refresh([]);
```
rohanmathur91: We are never resetting `syncTask`, which then does not syncs the upcoming local records

app/src/features/apiClient/helpers/modules/sync/localStore/components/AutoSyncLocalStoreDaemon.tsx
```diff
@@ -86,22 +96,26 @@ const LoggedOutDaemon: React.FC<{}> = () => {
 
     (async () => {
       try {
-        await syncTask;
-        const records = getAllRecords();
+        const result = await syncTask;
+        setSyncTask(null);
```
iostreamer-X, line 100: `setSyncTask` doesn't have a signature where it takes `null`. Also `syncTask` has type 'Promise | undefined`, let's either set undefined, or change the type of `syncTask` itself.

PR #3258, opened 2025-07-13
[ENGG-3862] feat: expand folder for matching search item

Aarushsr12:


app/src/features/apiClient/screens/apiClient/components/sidebar/components/collectionsList/CollectionsList.tsx
```diff
@@ -113,8 +113,26 @@ export const CollectionsList: React.FC<Props> = ({ onNewClick, recordTypeToBeCre
...
+          let parentId = childParentMap.get(record.collectionId);
+          while (parentId) {
+            recordsToExpand.add(parentId);
+            parentId = childParentMap.get(parentId);
+          }
+        }
+      });
+      setExpandedRecordIds((prev: string[]) => {
+        const newExpanded = [...prev, ...Array.from(recordsToExpand)];
+        return newExpanded;
```
RuntimeTerror10: Here the usage of spread operator and `Array.from()` can be avoided. Lets use an array instead of a Set and use `Array.concat()`. [refer](https://jonlinnell.co.uk/articles/spread-operator-performance)

RuntimeTerror10 (changes requested):

PR #3261, opened 2025-07-14
fix: api client global env sync from logged out state

rohanmathur91:


app/src/features/apiClient/helpers/modules/sync/localStore/store/syncServiceStore.ts
```diff
@@ -99,15 +100,33 @@ export const createSyncServiceStore = () => {
...
-      const envs = Object.values(allEnvs.data.environments);
-      const isJustGlobalEnv =
-        envs.length === 1 &&
-        envs[0].id === localStoreRepository.environmentVariablesRepository.getGlobalEnvironmentId();
+      const globalEnvId = localStoreRepository.environmentVariablesRepository.getGlobalEnvironmentId();
+      const globalEnv = allEnvs.data.environments[globalEnvId];
 
-      if (!isJustGlobalEnv) {
+      if (!globalEnv.variables || Object.keys(globalEnv.variables).length === 0) {
+        await localStoreRepository.environmentVariablesRepository.clear();
```
iostreamer-X: Why are we clearing the entire repository if there's no global env(or its variables)? Is there a case where other environments might exist at this point?

iostreamer-X (changes requested):

app/src/features/apiClient/helpers/modules/sync/localStore/store/syncServiceStore.ts
rohanmathur91: Missed it, updated, yes other envs can exist at that point.

iostreamer-X (approved):

PR #3264, opened 2025-07-16
chore: whitelist testing toolkit extension in manifest

wrongsahil:
Whiltelist Testing Toolkit extension in manifest so that it can communicate with Requestly Extension,

Extension is only present in chrome store right now. So only added in chrome manifest.


nafees87n (approved):
Looks good

PR #3265, opened 2025-07-16
[ENGG-3967]: show new webApp version available notification

rohanmathur91:


app/src/componentsV2/AppUpdateNotifier/analytics/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type NotificationType = "app_update";
+type ActionOnNotification = "close" | "app_refresh";
+
+export const trackAppUpdateNotificationViewed = (notificationType: NotificationType) => {
+  trackEvent(APP_NOTIFICATION.VIEWED, { notification_type: notificationType });
+};
+
+export const trackAppUpdateNotificationClicked = (notificationType: NotificationType, action: ActionOnNotification) => {
+  trackEvent(APP_NOTIFICATION.VIEWED, { notification_type: notificationType, action });
```
wrongsahil: `APP_NOTIFICATION.VIEWED` -> `APP_NOTIFICATION.CLICKED`

app/src/componentsV2/AppUpdateNotifier/AppUpdateNotifier.tsx
```diff
@@ -0,0 +1,103 @@
...
+            throw new Error("App data does not exist");
+          }
+
+          const currentAppVersion = process.env.REACT_APP_VERSION;
+
+          const appVersions: AppVersions = snapshot.val();
+          const latestAppVersion = appVersions.latestAppVersion;
+          const breakingAppVersion = appVersions.breakingAppVersion;
+
+          if (semver.valid(latestAppVersion) === null || semver.valid(breakingAppVersion) === null) {
```
wrongsahil: We shouldn't throw in this case for non existent version, we should treat them as `0.0.0` and let the rest of the code execute. 

Only in case of invalid version, we should captureException and treat it like `0.0.0` instead and let the rest of the code execute. This is so that any 1 (latestAppVersion or breakingAppVersion) invalid version doesn't impact other

app/src/componentsV2/AppUpdateNotifier/AppUpdateNotifier.tsx
```diff
@@ -0,0 +1,103 @@
...
+            setShowRefreshOption(true);
+            return;
+          }
+
+          // v1 <= v2: force refresh
+          if (semver.lte(currentAppVersion, breakingAppVersion)) {
+            trackAppUpdateForceReload(currentAppVersion, breakingAppVersion);
+            window.location.replace(window.location.href);
+            return;
+          }
```
wrongsahil: Above is really difficult to understand. How about this?

```
if(semver.lt(currentAppVersion, breakingAppVersion)) {
// Force Reload
} else if(semver.lt(currentAppVersion, latestAppVersion)) {
// Show App Update Notification
} else if(semver.eq(currentAppVersion, latestAppVersion)) {
// Do nothing
} else {
// Do nothing
}
```

wrongsahil: Also, it contains some edgecases. For eg.
- currentAppVersion > latestAppVersion, Then it will keep on showing update available notification. 

Also it is assuming latestAppVersion >= breakingAppVersion which might be right but we might accidentally not follow it which will lead to edgecase

app/src/componentsV2/AppUpdateNotifier/AppUpdateNotifier.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+          const currentAppVersion = process.env.REACT_APP_VERSION;
+
+          const appVersions: AppVersions = snapshot.val();
+          const latestAppVersion = appVersions.latestAppVersion;
+          const breakingAppVersion = appVersions.breakingAppVersion;
+
+          if (semver.valid(latestAppVersion) === null || semver.valid(breakingAppVersion) === null) {
+            throw new Error("Invalid app version format");
```
wrongsahil: Instead of throwing and capturing exception, can we just capture it here only and then return?

PR #3267, opened 2025-07-17
[ENGG-3980] handle autogenerated headers with placeholders

Aarushsr12:


app/src/features/apiClient/types.ts
```diff
@@ -21,7 +21,8 @@ export enum RequestMethod {
 export enum RequestContentType {
   RAW = "text/plain",
   JSON = "application/json",
-  FORM = "application/x-www-form-urlencoded",
+  //FORM = "application/x-www-form-urlencoded",
+  FORM = "multipart/form-data",
```
iostreamer-X: Why are we using on 'FORM'? There are two headers here 'form-encoded' and 'form-data', let's have them both, rather than using just 'form'.

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -118,10 +124,19 @@ export const createAutogeneratedStore = () => {
...
-    purgeAndAddHeaders: (namespace: AutogeneratedFieldsNamespace, params: Record<string, string>) => {
+    purgeAndAddHeaders: (
+      namespace: AutogeneratedFieldsNamespace,
+      params: Record<string, { value: string; prettyValue?: string }>
+    ) => {
       const { clearNamespace, _addHeaders } = get();
       clearNamespace(namespace, true);
-      _addHeaders(namespace, params);
+
+      const transformedParams: Record<string, string> = {};
```
iostreamer-X: We are losing information here, `Record<string, string>` can no longer be used to store headers. Let's not transform anything here, and update the underlying datastructure to take the new record(where we have prettyValue).

This function should not take any decisions about what should be accepted and what should be rejected.

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -60,7 +63,10 @@ export function parseAuth(
 export function parseContentType(contentType: RequestContentType): NamespaceContents {
   const queryParams: SimpleKeyValuePair[] = [];
   const headersMap = new Map<string, string>();
-  headersMap.set("content-type", contentType);
+
+  const contentTypeValue =
+    contentType === RequestContentType.FORM ? "multipart/form-data boundary<calculated at runtime>" : contentType;
```
iostreamer-X: Create a `const` for this prettyValue, rather than having this here as a string. Create a function which takes header and return {value, prettyValue}.

app/src/features/apiClient/screens/apiClient/components/clientView/APIClientView.tsx
```diff
@@ -238,8 +238,11 @@ const APIClientView: React.FC<Props> = ({
         contentTypeHeader.key = CONTENT_TYPE_HEADER;
         contentTypeHeader.value = contentType;
 
+        const prettyValue =
+          contentType === RequestContentType.FORM ? "multipart/form-data boundary<calculated at runtime>" : undefined;
```
iostreamer-X: Introduce a new header enum dedicatedly for form-data, and use the function mentioned in the following components to set the prettyValue.

iostreamer-X (changes requested):

app/src/features/apiClient/types.ts
Aarushsr12: yup, just for testing adding string here since not having the radio btn for multipart form as of now

Aarushsr12: when implement design will make the change

app/src/features/apiClient/screens/apiClient/components/clientView/APIClientView.tsx
```diff
@@ -239,12 +240,12 @@ const APIClientView: React.FC<Props> = ({
         contentTypeHeader.value = contentType;
 
         purgeAndAddHeaders(AutogeneratedFieldsNamespace.CONTENT_TYPE, {
-          [contentTypeHeader.key]: contentTypeHeader.value,
+          [contentTypeHeader.key]: getResolvedHeaders(contentType),
```
nsrCodes: suggestion:
```suggestion
          [contentTypeHeader.key]: resolveHeaders(contentType),
```

app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -116,7 +116,7 @@ export const sanitizeEntry = (entry: RQAPI.Entry, removeDisabledKeys = true) =>
   if (entry.request.body != null) {
     if (!supportsRequestBody(entry.request.method)) {
       sanitizedEntry.request.body = null;
-    } else if (entry.request.contentType === RequestContentType.FORM) {
+    } else if (entry.request.contentType === RequestContentType.urlEncodedForm) {
       sanitizedEntry.request.body = sanitizeKeyValuePairs(
         entry.request.body as RQAPI.RequestFormBody,
```
nsrCodes: should we make a function like this:

```
const isRequestWithURLEncodedForm = (request: RQAPI.Request): request is RQAPI.Request & { 
  contentType: RequestContentType.URL_ENCODED_FORM;
  body: RQAPI.RequestFormBody;
} => {
  return request.contentType === RequestContentType.URL_ENCODED_FORM;
};
```
because:
1. removes the type assertion need on L#121
2. consolidates the check into a function that can be used elsewhere -> and makes it clear what the check is for a `entry.request` to be a `urlEncodedForm`

nsrCodes: if we agree to do this, it also makes sense to have a similar function for multi-part form requests. 
that would fill the gap of the implementation [here](https://github.com/requestly/requestly-desktop-app/pull/193) _(and also for the extension implementation in the future)_

> Note: these changes (and my comment) do not seem related to the (perceived) scope of this PR.

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -158,7 +180,24 @@ export const createAutogeneratedStore = () => {
       _addQueryparams(namespace, queryParams);
     },
 
-    getAllHeaders() {
+    getHeaderWithPrettyValue() {
```
nsrCodes:
```suggestion
    getHeadersWithPrettyValue() {
```

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -6,10 +6,27 @@ export enum AutogeneratedFieldsNamespace {
   AUTH = "auth",
   CONTENT_TYPE = "content_type",
 }
+export enum PrettifiedHeaders {
+  MULTI_PART_FORM = "multipart/form-data; boundary=<calculated when request is sent>",
+}
+interface HeaderValues {
+  originalValue: string;
+  prettyValue?: string;
+}
```
nsrCodes: could be made cleaner. one possible way:
```suggestion
interface HeaderEntry {
  value: string;
  prettyValue?: string; // could also use the name `placeHolder`
}
```

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -41,9 +59,9 @@ export function parseAuth(
   // ---> now here we can use that function already defined since it has the same signature of params
   const { headers, queryParams } = extractAuthHeadersAndParams(resolver ? resolver(auth) : auth);
   //convert the headers from KP to map
-  const headersMap = new Map<string, string>();
+  const headersMap = new Map<string, { originalValue: string; prettyValue?: string }>();
```
nsrCodes: why are we not using the new interface here?

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -167,7 +206,7 @@ export const createAutogeneratedStore = () => {
           const [key, value] = headerEntry;
           result.push({
             key,
-            value,
+            value: value.originalValue,
```
nsrCodes: this is confusing and hints at a problem with the names and abstractions. 

if code is becoming along the lines of `stuff.realStuff` this will lead to code like:
1. `setStuff({realStuff: arg})` which is what seems to have appeared in [L#62](https://github.com/requestly/requestly/blob/1ee2584cfce45907baf3aaa9a011708c790e172a/app/src/features/apiClient/store/autogenerateStore.ts#L64)
2. confusing checks like `isRealStuff(stuff)` i.e. we might end up having checks like `isOriginalValue(header.value)` which would lead to so much confusion in any other file except here

This should be solved by renaming the new interface. mentioned a suggestion for it in the comments above

app/src/features/apiClient/types.ts
```diff
@@ -98,7 +99,8 @@ export namespace RQAPI {
 
   export type RequestBodyContainer = {
     text?: string;
-    form?: KeyValuePair[];
+    formURLEncoded?: KeyValuePair[];
+    multipartForm?: KeyValuePair[];
```
nsrCodes: [`RequestBodyStateManager`](https://github.com/requestly/requestly/blob/fb929d4f1496ef67828a3556ace29a0366a46791/app/src/features/apiClient/screens/apiClient/components/clientView/components/request/request-body-state-manager.ts#L13-L41) has not been updated to work with these changes to the type

nsrCodes: btw if the scope of this is to add support for showing pretty headers, why are there partial changes for introducing multipart form?

do we need to change the base of this PR to make it cleaner?

app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -504,9 +504,9 @@ export const apiRequestToHarRequestAdapter = (apiRequest: RQAPI.Request): HarReq
         mimeType: RequestContentType.JSON,
         text: apiRequest.body as string,
       };
-    } else if (apiRequest?.contentType === RequestContentType.FORM) {
+    } else if (apiRequest?.contentType === RequestContentType.urlEncodedForm) {
       harRequest.postData = {
-        mimeType: RequestContentType.FORM,
+        mimeType: RequestContentType.urlEncodedForm,
```
nsrCodes: does not handle the `multipartForm` case

> Note: these changes (and my comment) do not seem related to the (perceived) scope of this PR.

app/src/features/apiClient/helpers/apiClientExecutor/apiClientExecutor.ts
```diff
@@ -73,8 +73,8 @@ export class ApiClientExecutor {
 
     this.renderedVariables = renderedVariables;
 
-    const { getAllHeaders, getAllQueryParams } = this.autoGeneratedStore.getState();
-    const headers = getAllHeaders().map((kv, index) => ({ ...kv, id: index, isEnabled: true }));
+    const { getHeaderWithOriginalValue, getAllQueryParams } = this.autoGeneratedStore.getState();
+    const headers = getHeaderWithOriginalValue().map((kv, index) => ({ ...kv, id: index, isEnabled: true }));
```
nsrCodes:
```suggestion
    const { getHeadersWithOriginalValue, getAllQueryParams } = this.autoGeneratedStore.getState();
    const headers = getHeadersWithOriginalValue().map((kv, index) => ({ ...kv, id: index, isEnabled: true }));
```

nsrCodes (changes requested):
These changes contain incomplete changes that introduce some version of the types and logic for multi part form support. I believe that isn't in the scope of this PR 

I haven't added comments for all such cases, but I suggest to coordinate with @nafees87n to get a common base branch for this PR and https://github.com/requestly/requestly/pull/3270 so that:

a. these don't cause merge conflicts later
b. makes sure that both are on the same page with what the final types will be
c. makes it easier to review isolated PRs like this one

app/src/features/apiClient/helpers/apiClientExecutor/apiClientExecutor.ts
iostreamer-X: How about we keep it as its now(`getAllheaders`), but while rendering we use a different function. This is to mark that using `value` is the invariant, and using prettyValue is the special case.

Current structure makes it look vice versa.

app/src/features/apiClient/types.ts
Aarushsr12: Just added here, for testing the multipartform autogenerated headers, removing these partial changes of multipart form, will go with https://github.com/requestly/requestly/pull/3270 base changes

app/src/features/apiClient/screens/apiClient/utils.ts
Aarushsr12: this is again for the radio btn usecase mentioned reason above, can remove this for now & use changes coming from base branch

app/src/features/apiClient/helpers/apiClientExecutor/apiClientExecutor.ts
Aarushsr12: - bring back the `getAllHeaders` function here, dealing with `value`
- seperate function for prettyValue special case, which will be used in autogenerated table component ?

this is the ask?

iostreamer-X: yep yep

app/src/features/apiClient/screens/apiClient/utils.ts
Aarushsr12: - Noting this, In UI changes branch will address this issue

app/src/features/apiClient/store/autogenerateStore.ts
Aarushsr12: Have a doubt related to this

- `HeaderEntry` is mainly for the autogenerated store interface, methods defined 
using this for a variable created inside a function L#62 which is headersMap here, is the right way?

1. headersMap structure can be changed in future, HeaderEntry might make it bit coupled?
2. But one thing is that it is just breaking down the `params` which is using HeaderEntry, respecting this we can use HeaderEntry with `headersMap`

app/src/features/apiClient/store/autogenerateStore.ts
Aarushsr12: With suggested interface will might face bit confusion:
```
interface HeaderEntry {
  value: string;
  prettyValue?: string; // could also use the name `placeHolder`
}
```
issue seems with `value` field here, when writing this

```
for (const headerEntry of headers.entries()) {
          const [key, value] = headerEntry;
          result.push({
            key,
            value: value.value
          });
        }
```

Aarushsr12: this might be confusing, but scope of this issue will remain here only, the originalValue usage over the next changes won't become confusing if we simply use `value` field only

app/src/features/apiClient/store/autogenerateStore.ts
Aarushsr12: We should move with placeholder, sounds more apt & technical , prettyValue sounds not related

app/src/features/apiClient/store/autogenerateStore.ts
iostreamer-X: The coupling IS the point here, we want to change how we store and use headers fundamentally.

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -41,9 +59,9 @@ export function parseAuth(
   // ---> now here we can use that function already defined since it has the same signature of params
   const { headers, queryParams } = extractAuthHeadersAndParams(resolver ? resolver(auth) : auth);
   //convert the headers from KP to map
-  const headersMap = new Map<string, string>();
+  const headersMap = new Map<string, { originalValue: string; prettyValue?: string }>();
   headers.forEach(({ key, value }) => {
-    headersMap.set(key, value);
+    headersMap.set(key, { originalValue: value });
```
iostreamer-X: Doesn't this mean that, when refreshed, there's no placeholder and the table would show the actual value? As you can see when parsing auth, which happens on load, we are not setting any placeholder.

Please check this: https://github.com/requestly/requestly/pull/3267#discussion_r2212211571

app/src/features/apiClient/store/autogenerateStore.ts
iostreamer-X: Can go with placeholder.

app/src/features/apiClient/store/autogenerateStore.ts
iostreamer-X: @Aarushsr12 based on our discussion, the new function(which returns headers entry) would need to be called here, as well as any place that initializes headers in autogen store.

app/src/features/apiClient/constants.ts
```diff
@@ -73,3 +73,7 @@ export const DEFAULT_SCRIPT_VALUES = {
  * Also checks the braces, cases where ex:{{RQ_CLIENT_ID}} environment variables are used, this should be not marked invalid other than this will be flagged
  */
 export const INVALID_KEY_CHARACTERS = /^(?!\{\{.*\}\}|^\{\{?$|^\}\}?$)[^!#$%&'*+\-.^_`|~0-9A-Za-z]/;
+
+export enum HeadersType {
```
iostreamer-X: This could be just `Headers`

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -1,32 +1,47 @@
...
-  originalValue: string;
-  prettyValue?: string;
+interface HeaderEntry {
+  value: string;
+  placeHolder?: string;
 }
 
-export const getResolvedHeaders = (contentType: RequestContentType): HeaderValues => {
-  switch (contentType) {
+const resolveContentTypeHeaders = (value: string): HeaderEntry => {
```
iostreamer-X: Let's not call this resolve, it has a different connotation. This is more like `createHeaderEntryFrom_`

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -1,32 +1,47 @@
...
+      return {
+        value: value,
+      };
+  }
+};
+
+/*
+create resolver functions for headers to add
+*/
+export const resolveHeaders = (key: string, value: string): HeaderEntry => {
```
iostreamer-X: Same here, this is just `createHeaderEntry`

app/src/features/apiClient/store/autogenerateStore.ts
```diff
@@ -36,17 +51,17 @@ export interface AutogeneratedFieldsStore {
...
   //Header methods
   initialize: (namespaces: AutogeneratedFieldsStore["namespaces"]) => void;
-  _addHeaders: (namespace: AutogeneratedFieldsNamespace, params: Record<string, HeaderValues>) => void;
+  _addHeaders: (namespace: AutogeneratedFieldsNamespace, params: Record<string, HeaderEntry>) => void;
   clearNamespace: (namespace: AutogeneratedFieldsNamespace, skipVersionUpdate?: boolean) => void;
-  purgeAndAddHeaders: (namespace: AutogeneratedFieldsNamespace, params: Record<string, HeaderValues>) => void;
-  getHeaderWithPrettyValue: () => SimpleKeyValuePair[];
-  getHeaderWithOriginalValue: () => SimpleKeyValuePair[];
+  purgeAndAddHeaders: (namespace: AutogeneratedFieldsNamespace, params: Record<string, HeaderEntry>) => void;
+  getHeadersWithPrettyValue: () => SimpleKeyValuePair[];
```
iostreamer-X: If we are using `placeholder`, should this be called something else? Like `getPlaceholderHeaders`?

app/src/features/apiClient/screens/apiClient/components/clientView/APIClientView.tsx
```diff
@@ -239,7 +240,7 @@ const APIClientView: React.FC<Props> = ({
         contentTypeHeader.value = contentType;
 
         purgeAndAddHeaders(AutogeneratedFieldsNamespace.CONTENT_TYPE, {
-          [contentTypeHeader.key]: contentTypeHeader.value,
+          [contentTypeHeader.key]: resolveHeaders(contentTypeHeader.key, contentTypeHeader.value),
```
iostreamer-X: Do we still have a `resolveHeaders`? I couldn't find in the autogenerate stores definition.

app/src/features/apiClient/store/autogenerateStore.ts
Aarushsr12: Yup, it got missed, now fixed

app/src/features/apiClient/screens/apiClient/components/clientView/APIClientView.tsx
Aarushsr12: Yup, missed here, now fixed

iostreamer-X (approved):

PR #3285, opened 2025-07-24
[ENGG-4010]: Create a factory for different request views in API Client

RuntimeTerror10:


app/src/features/apiClient/screens/apiClient/clientView/ClientViewFactory.tsx
```diff
@@ -0,0 +1,37 @@
...
+    return <GraphQLClientViewContainer />;
+  }
+
+  return (
+    <HttpClientViewContainer
+      selectedEntryDetails={apiRecord as RQAPI.HttpApiRecord}
+      handleAppRequestFinished={handleRequestFinished}
+      onSaveCallback={onSaveCallback}
+      isCreateMode={isCreateMode}
+    />
```
wrongsahil, line 35: For none of the above, we should raise exception

wrongsahil (approved):

app/src/features/apiClient/screens/apiClient/clientView/ClientViewFactory.tsx
RuntimeTerror10: Shouldn't we keep a fallback component for the existing HTTP requests which which don't have `type` inside `data`?

wrongsahil: If entries without type are HTTP Entries only, then it's fine. But if it is a non deterministic state, then we should raise

PR #3280, opened 2025-07-23
[ENGG-3988]: create stores for each logical unit for environments 

rohanmathur91:

PR #3288, opened 2025-07-24
refactor: Stores aggregator for apiClient

nsrCodes:
builds on top of https://github.com/requestly/requestly/pull/3280 but still expects the following changes there:
- a store for errored recoreds
- updates to the `createEnvironmentStore` definition to allow passing global records


nsrCodes:
superseded by https://github.com/requestly/requestly/pull/3294

PR #3316, opened 2025-07-29
feat: wrapper for all stores and repositories to be easily accessible

nsrCodes:


app/src/features/apiClient/contexts/meta.tsx
```diff
@@ -0,0 +1,15 @@
+import { useContext } from "react";
+import { ApiClientRepositoryInterface } from "../helpers/modules/sync/interfaces";
+import { useApiClientRepository } from "../helpers/modules/sync/useApiClientSyncRepo";
+import { AllApiClientStores, ApiRecordsStoreContext } from "../store/apiRecords/ApiRecordsContextProvider";
+
+export type ApiClientFeatureContext = {
+  stores: AllApiClientStores | null;
```
rohanmathur91: This wont be `null` either whole context value is `null | ApiClientFeatureContext` right?

PR #3352, opened 2025-08-06
ENGG-4101: dropdown in query builder to add operations

RuntimeTerror10:
image


app/src/features/apiClient/screens/apiClient/components/views/graphql/components/SchemaBuilder/SchemaBuilder.tsx
nafees87n, line 1: Can we hide subscription from the drop down as well?

PR #3359, opened 2025-08-08
feat: variable store for runtime variable

nsrCodes:


app/src/features/apiClient/store/variables/runtimeVariables.store.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    delete(key) {
+      const { data } = get();
+
+      if (!data.has(key)) {
+        return;
+      }
+
+      data.delete(key);
+      set({ data });
```
iostreamer-X: Check with @rohanmathur91 on using immer and removing version

PR #3364, opened 2025-08-11
[ENGG-4106]: stores for multi workspace view

rohanmathur91:
- Workspaces store
- Contexts store

### DIFF
--- a/app/src/features/apiClient/store/apiClientFeatureContext/apiClientFeatureContext.store.ts
+++ b/app/src/features/apiClient/store/apiClientFeatureContext/apiClientFeatureContext.store.ts
@@ -5,17 +5,15 @@ import { create, useStore } from "zustand";
 import { NativeError } from "errors/NativeError";
 import { useShallow } from "zustand/shallow";
 
-type ApiClientFeatureContext = {
+export type ApiClientFeatureContext = {
   id: RenderableWorkspaceState["id"]; // Instead of id, should this be workspaceId or as a new property ?
   stores: AllApiClientStores;
   repositories: ApiClientRepositoryInterface;
 };
 
 type ApiClientFeatureContextProviderState = {
-  version: number;
   contexts: Map<RenderableWorkspaceState["id"], ApiClientFeatureContext>;
 
-  _incrementVersion(): void;
   addContext(context: ApiClientFeatureContext): void;
   removeContext(id: RenderableWorkspaceState["id"]): void;
   getContext(id: RenderableWorkspaceState["id"]): ApiClientFeatureContext | undefined;
@@ -26,18 +24,16 @@ type ApiClientFeatureContextProviderState = {
 function createApiClientFeatureContextProviderStore() {
   return create<ApiClientFeatureContextProviderState>()((set, get) => {
     return {
-      version: 0,
-      contexts: new Map<RenderableWorkspaceState["id"], ApiClientFeatureContext>(),
+      contexts: new Map(),
 
       addContext(context) {
-        const { contexts, _incrementVersion } = get();
+        const { contexts } = get();
         contexts.set(context.id, context);
         set({ contexts });
-        _incrementVersion();
       },
 
       removeContext(id) {
-        const { contexts, _incrementVersion } = get();
+        const { contexts } = get();
         const context = contexts.get(id);
 
         if (!context) {
@@ -46,7 +42,6 @@ function createApiClientFeatureContextProviderStore() {
 
         contexts.delete(id);
         set({ contexts });
-        _incrementVersion();
       },
 
       getContext(id) {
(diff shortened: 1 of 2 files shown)


app/src/features/apiClient/store/apiClientFeatureContext/apiClientFeatureContext.store.ts
```diff
@@ -0,0 +1,88 @@
...
+      },
+
+      getContext(id) {
+        const { contexts } = get();
+        return contexts.get(id);
+      },
+
+      getSingleViewContext() {
+        const { contexts } = get();
+        const context = Object.values(Object.fromEntries(contexts))[0];
```
iostreamer-X: No need to change types, can be simply done by `contexts.values().next().value`

app/src/features/apiClient/store/apiClientFeatureContext/apiClientFeatureContext.store.ts
```diff
@@ -0,0 +1,88 @@
...
+
+        if (!context) {
+          throw new NativeError("Context does not exist in single mode");
+        }
+
+        return Object.values(Object.fromEntries(contexts))[0];
```
iostreamer-X: No need to convert types again.

app/src/features/apiClient/store/apiClientFeatureContext/apiClientFeatureContext.store.ts
```diff
@@ -0,0 +1,88 @@
...
+  removeContext(id: RenderableWorkspaceState["id"]): void;
+  getContext(id: RenderableWorkspaceState["id"]): ApiClientFeatureContext | undefined;
+  getSingleViewContext(): ApiClientFeatureContext;
+  refreshContext(id: RenderableWorkspaceState["id"], context: ApiClientFeatureContext): void;
+};
+
+function createApiClientFeatureContextProviderStore() {
+  return create<ApiClientFeatureContextProviderState>()((set, get) => {
+    return {
+      version: 0,
```
iostreamer-X: Let's remove version here. We have used it as a crutch when there are changes that can not be reacted upon, here we have only field `contexts` and not a deeply nested structure. Putting `version` is extra bookkeeping that I don't see will provide any value in future as well.

app/src/features/apiClient/store/apiClientFeatureContext/apiClientFeatureContext.store.ts
```diff
@@ -0,0 +1,88 @@
+import { ApiClientRepositoryInterface } from "features/apiClient/helpers/modules/sync/interfaces";
+import { AllApiClientStores } from "../apiRecords/ApiRecordsContextProvider";
+import { RenderableWorkspaceState } from "../multiWorkspaceView/multiWorkspaceView.store";
+import { create, useStore } from "zustand";
+import { NativeError } from "errors/NativeError";
+import { useShallow } from "zustand/shallow";
+
+type ApiClientFeatureContext = {
```
iostreamer-X: Please export all types. Remove existing types if there's conflict.

app/src/features/apiClient/store/multiWorkspaceView/multiWorkspaceView.store.ts
```diff
@@ -0,0 +1,91 @@
...
+        set({ state });
+      },
+    };
+  });
+}
+
+function createApiClientMultiWorkspaceViewStore() {
+  return create<ApiClientMultiWorkspaceViewState>()((set, get) => {
+    return {
+      viewMode: get().selectedWorkspaces.length > 0 ? ApiClientViewMode.MULTI : ApiClientViewMode.SINGLE,
```
iostreamer-X: This is not a getter field, so this is being initialised just at the beginning but never updated.

iostreamer-X: Need to have a dedicated method to set the view mode as well.

iostreamer-X: Also selectedWorkspaces.length  is not a good indicator of being in multi or single, as there could be an interim state where we haven't selected any workspaces for our view but are about to. Let's not any heuristics here and provide only the functions. The commands will have this "business" logic and accordingly set the mode here.

app/src/features/apiClient/store/multiWorkspaceView/multiWorkspaceView.store.ts
```diff
@@ -0,0 +1,91 @@
...
+        const { selectedWorkspaces } = get();
+        set({ selectedWorkspaces: [...selectedWorkspaces, createRenderableWorkspaceStore(workspace)] });
+      },
+
+      removeWorkspace(id) {
+        const { selectedWorkspaces } = get();
+
+        const workspace = selectedWorkspaces.find((w) => w.getState().id === id);
+        if (!workspace) {
+          throw new NativeError("Workspace not found!").addContext({ workspaceId: id });
```
iostreamer-X: Let's keep deletes idempotent, if the workspace is already deleted, then all good!

iostreamer-X (changes requested):

PR #3371, opened 2025-08-11
feat: commands for setting up context and view

nsrCodes:


app/src/features/apiClient/commands/multiView/addWorkspaceToView.command.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+  });
+};
+
+export const addWorkspaceToView = async (param: Workspace) => {
+  if (param.type !== WorkspaceType.LOCAL)
+    throw new NativeError("[ADD TO VIEW] Multi view only avaiable for local workspaces");
+  const { id } = await addWorkspaceToAppCtx(param);
+
+  const renderableParam = {
```
iostreamer-X, line 23: This needs to change where we add to `ApiClientMultiWorkspaceViewState` first but with state=loading and then mark it as either errored or loaded after context is setup.

nsrCodes:
changes got merged in as part of https://github.com/requestly/requestly/pull/3381

PR #3411, opened 2025-08-19
[ENGG-4709] UI: Variable popover

Aarushsr12:
https://github.com/user-attachments/assets/323141b2-7db5-4411-80e2-0d2a0af77172


rohanmathur91 (changes requested):
Alignment issue:

Image

Popup keeps growing in width:

Image

PR #3412, opened 2025-08-19
[ENGG-4245] chore: created extension constants in shared module

nafees87n:
1. Moved common extension types like messages, storage_keys to requestly/shared so that UI and extension have a common source
2. Not all types have been moved. Only some types are moved to keep the testing scope very limited.
3. To migrate all the types, we need to come up with a plan to efficiently test

## Summary by CodeRabbit

- Refactor
  - Centralized extension messages and storage keys into shared constants for consistent usage across the app.
  - Standardized message handling for cURL import and session replay checks without changing behavior.

- Chores
  - Upgraded TypeScript and build tooling for improved compatibility and stability.
  - Updated bundler configuration and module resolution to align with modern tooling.
  - Added shared package as a development dependency to streamline constant sharing.

No user-facing functionality changes.


shared/src/types/common/index.ts
```diff
@@ -1,2 +1,3 @@
 export * as network from "./network";
 export * from "./email";
+export * as ExtensionConstants from "./extension";
```
wrongsahil, line 3:
```suggestion
export * as extension from "./extension";
```

For consistency, lets export it as extension only

shared/src/types/common/extension/index.ts
```diff
@@ -0,0 +1,9 @@
+export enum StorageKeys {
+  POPUP_CONFIG = "popup_config",
+}
+
+export enum ExtensionMessages {
+  OPEN_CURL_IMPORT_MODAL = "openCurlImportModal",
+  IS_SESSION_REPLAY_ENABLED = "isSessionReplayEnabled",
+  TRIGGER_OPEN_CURL_MODAL = "triggerOpenCurlModal",
```
wrongsahil, line 8: This should be on ClientMessages

app/src/hooks/useInitPopupConfig.ts
```diff
@@ -17,7 +18,7 @@ export const useInitPopupConfig = () => {
   ]);
 
   const fetchConfig = useCallback(async () => {
-    const popupConfig = await StorageService(appMode).getRecord("popup_config");
+    const popupConfig = await StorageService(appMode).getRecord(ExtensionConstants.StorageKeys.POPUP_CONFIG);
```
wrongsahil, line 21: We should define the type for popupConfig too in `@requestly/shared`

wrongsahil (changes requested):

PR #3470, opened 2025-08-29
fix: set user role as admin for local workspaces

RuntimeTerror10:
## Summary by CodeRabbit

- New Features
  - Local workspaces now grant full admin permissions by default, enabling full management without account checks.
- Bug Fixes
  - Improved role detection to handle edge cases more gracefully, reducing unexpected permission issues.
  - Ensures users without authentication or workspaces lacking an ID are treated as admins when appropriate, maintaining uninterrupted access.


nafees87n (approved):
LGTM

PR #3499, opened 2025-09-04
[ENGG-4187]: Add support for opening collection/workspace folder path in file explorer

RuntimeTerror10:
https://github.com/user-attachments/assets/577b2728-d17b-4062-995e-a41b64b78870

## Summary by CodeRabbit

- New Features
  - Multi-workspace aware breadcrumbs appear when local sync is enabled, adapting between single and multi-view modes.
  - Breadcrumbs are clickable in supported contexts, letting you open corresponding folders in your file manager.
  - Tooltips show the full active workspace path; breadcrumb labels are truncated for readability.
  - Design system breadcrumb now supports per-crumb click handlers for richer interactions.

- Chores
  - Added feature flag and compatibility entry for "open_path_in_file_manager".


app/src/features/apiClient/screens/apiClient/components/views/components/ApiClientBreadCrumb/ApiClientBreadCrumb.tsx
```diff
@@ -29,42 +33,62 @@ export const BreadcrumbType = {
...
-        return {
-          label: getData(id).name,
-          pathname: "",
-          isEditable: false,
-        };
+  const handleBreadcrumbLabelClick = useCallback(
+    (index: number) => {
+      const pathItems = [currentWorkspacePath, ...parentCollections.map((item) => item.name)];
+      const pathToOpen = pathItems.slice(0, index + 1).join("/");
+      window.RQ.DESKTOP.SERVICES.IPC.invokeEventInMain("open-path-in-file-manager", {
```
nsrCodes: I know this isn't relevant for now, but this will crash if this component is loaded in web. can we please optional chain here.

nsrCodes: will also need to add a compatibility check here otherwise this will break in older versions (as the handler for the invoke isn't present there)

PR #3501, opened 2025-09-04
fix: don't show connected screen on specific routes

nsrCodes:


app/src/hooks/useDesktopAppConnection.ts
```diff
@@ -6,6 +6,25 @@ import PSMH from "../config/PageScriptMessageHandler";
 import { CONSTANTS as GLOBAL_CONSTANTS } from "@requestly/requestly-core";
 import Logger from "lib/logger";
 
+import PATHS from "config/constants/sub/paths";
+
+const EXCLUDED_PATHS = [
+  PATHS.AUTH.SIGN_IN.RELATIVE,
+  PATHS.AUTH.SIGN_UP.RELATIVE,
+  PATHS.AUTH.DEKSTOP_SIGN_IN.RELATIVE,
```
RuntimeTerror10, line 14: Typo `DESKTOP`

RuntimeTerror10 (approved):

app/src/hooks/useDesktopAppConnection.ts
nsrCodes: seems like a part of the app now 😄 
keeping it for now, but surely fix adhoc

PR #3560, opened 2025-09-22
added generic way to create apiClientTreeBus

nafees87n:


app/src/features/apiClient/helpers/apiClientTreeBus/createApiClientTreeBus.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useMemo, useState } from "react";
+import { useApiClientFeatureContext } from "../../contexts/meta";
+import { Subscription } from "./subscription";
+import { ApiClientEventTopic } from "./types";
+
+export const createApiClientTreeBus = <K extends ApiClientEventTopic, TData>(
+  topic: K,
+  dataFetcher: (ctx: any, ...args: any[]) => TData
```
iostreamer-X, line 8: 1. Let's call this handler
2. This doesn't seem to handle types e2e, for example if topic is `K` then the handle must take a certain kind of parameter

iostreamer-X (changes requested):

PR #3573, opened 2025-09-24
Feature/stable ids local workspace

iostreamer-X:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error toasts and state resets when creating records/environments; close corrupted tabs after refresh; more reliable context refresh guard and per-root locking to reduce races.

* **Refactor**
  * Simplified cross-workspace record moves and removed unnecessary forced refreshes; reduced tab-closing side-effects in collections UI; memoized query-params store.

* **New Features**
  * UUID-based IDs for records/collections and records now include collection association; environments support a global flag with improved global handling; compatibility gating with a mandatory update screen.

* **UI**
  * Breadcrumb label shortened to remove workspace name; added styles and optional standalone rendering for mandatory update screen.


app/src/features/apiClient/commands/records/moveRecordsAcrossWorkspace.command.ts
```diff
@@ -20,77 +16,17 @@ export async function moveRecordsAcrossWorkspace(
...
-    const newId = isApiCollection(record)
-      ? apiClientRecordsRepository.generateCollectionId(record.name, parentId)
-      : apiClientRecordsRepository.generateApiRecordId(parentId);
-
-    oldToNewIdMap.set(record.id, newId);
-    newToOldIdMap.set(newId, record.id);
+  const result = await moveRecords(ctx, {
+    recordsToMove: recordsToMove,
+    collectionId: destination.collectionId,
   });
```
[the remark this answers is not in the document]
iostreamer-X: This is handled on desktop app side which takes care of this automatically. The fs layer doesn't care about different workspaces and maintains a singleton mapping of ids and file paths. Which is why special workspace code is not needed.

And if the source workspace and destination workspace is indeed different, we force refresh which re-syncs everything from scratch.

app/src/services/fsManagerServiceAdapter.ts
```diff
@@ -178,13 +179,21 @@ export class FsManagerServiceAdapter extends BackgroundServiceAdapter {
...
+    let lock = this.lockMap.get(rootPath);
+    if(!lock) {
+      lock = withTimeout(new Mutex(), 10 * 1000);
+      this.lockMap.set(rootPath, lock);
+    }
+    await lock.acquire();
     if (this.cache.has(rootPath)) {
       console.log("got provider from cache");
+      lock.release();
       return this.cache.get(rootPath);
```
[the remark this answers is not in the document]
iostreamer-X: There actually is a `release` method.
https://github.com/DirtyHairy/async-mutex?tab=readme-ov-file#unscoped-release

app/src/features/apiClient/helpers/modules/sync/local/services/LocalApiClientEnvSync.ts
```diff
@@ -2,22 +2,33 @@ import { EnvironmentData, EnvironmentMap } from "backend/environment/types";
...
 
   private async getAdapter() {
     return fsManagerServiceAdapterProvider.get(this.meta.rootPath);
   }
 
+  private setGlobalId(id: string) {
+    if(!this.globalId) {
+      this.globalId = id;
+    }
+  }
```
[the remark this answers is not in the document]
iostreamer-X: On disk global id deletion or uuid change count as catastrophic failure as the fs state once created CAN NOT change for any reason.

app/src/features/apiClient/screens/apiClient/components/views/http/HttpClientView.tsx
```diff
@@ -819,13 +819,15 @@ const HttpClientView: React.FC<Props> = ({
 
 const WithQueryParamsProvider = (Component: React.ComponentType<any>) => {
   return (props: any) => {
+    const record = useAPIRecordsStore().getState().getData(props.apiEntryDetails.id) as RQAPI.ApiClientRecord;
+    const entry = record?.data as RQAPI.HttpApiEntry || props.apiEntryDetails.data;
```
RuntimeTerror10, line 822: React does not allow usage of hooks inside callbacks functions.

RuntimeTerror10:
```js
const WithQueryParamsProvider = (Component: React.ComponentType<any>) => {
  const WrapperViewComponent = (props: any) => {
    const record = useAPIRecordsStore().getState().getData(props.apiEntryDetails.id) as RQAPI.ApiClientRecord;
    const entry = (record?.data as RQAPI.HttpApiEntry) || props.apiEntryDetails.data;
    return (
      <ErrorBoundary>
        <PathVariablesProvider pathVariables={entry.request?.pathVariables}>
          <QueryParamsProvider entry={entry}>
            <Component {...props} />
          </QueryParamsProvider>
        </PathVariablesProvider>
      </ErrorBoundary>
    );
  };
  return WrapperViewComponent;
};
```

RuntimeTerror10: Also should we do this for GraphQLView? It reads record from `useApiRecordState` and not `useAPIRecordsStore `

iostreamer-X: > React does not allow usage of hooks inside callbacks functions.

That's actually a component, the signature is (props) => JSX.

> Also should we do this for GraphQLView? It reads record from useApiRecordState and not useAPIRecordsStore 

That's actually the correct way, which is take id and then reactively get the record. Ideally this is how it should be in `HttpClientView`

PR #3575, opened 2025-09-24
[ENGG-4417]: Add path variables support in HTTP client

RuntimeTerror10:
## Summary by CodeRabbit

- New Features
  - Path variables support in the HTTP client: auto-detect from URLs, edit and persist values, include them in saved/sent requests, and compile URLs with provided values before sending.
  - Added a context/store and hook plus a utility to extract variable keys from URLs.

- UI
  - Renamed “Query Params” tab to “Params” and added a Path Variables table alongside query params with a title.

- Style
  - Added styling for the Path Variables table and params header.


RuntimeTerror10:
@coderabbitai review

RuntimeTerror10:
@coderabbitai review

app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -767,3 +767,25 @@ export const checkForLargeFiles = (body: RQAPI.RequestBody): boolean => {
...
 };
+
+export const extractPathVariablesFromUrl = (url: string) => {
+  if (!url) {
+    return [];
+  }
+
+  const urlWithScheme = addUrlSchemeIfMissing(url);
+  const pathname = new URL(urlWithScheme).pathname;
+
```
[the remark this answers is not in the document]
RuntimeTerror10: new URL() does not throw for the example input you shared. also `http:///users` will be part of origin so it is expected that no path param will be matched there

RuntimeTerror10:
@coderabbitai review

app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -767,3 +767,25 @@ export const checkForLargeFiles = (body: RQAPI.RequestBody): boolean => {
...
+  const variablePattern = /(?<!:):(\w+)/g;
```
nsrCodes: this means that the key cannot have non-word characters like `-` or `.`

fine for now, but should be mentioned in the documentation

app/src/features/apiClient/screens/apiClient/components/views/http/components/PathVariableTable/PathVariableTableRow.tsx
```diff
@@ -0,0 +1,94 @@
...
+interface EditableCellProps {
+  title: React.ReactNode;
+  editable: boolean;
+  dataIndex: keyof RQAPI.PathVariable;
+  record: RQAPI.PathVariable;
+  environmentVariables: ScopedVariables;
+  handleUpdateVariable: (record: RQAPI.PathVariable) => void;
+}
+
+export const EditableCell: React.FC<React.PropsWithChildren<EditableCellProps>> = ({
```
nsrCodes: this is very close to the existing EditableCell component of key value table.

Differences - A missing checkbox and an extra column for description

Since we know we'll eventually be moving to that after this thoroughly goes through product, either 
a. put in the effort to make the original `EditableCell` and `KeyValueTable` component extendable to cater this (and existing) usecases, or
b. rename the components here to avoid confusion and add a todo to revist and change this once the experience here has been finalized

app/src/features/apiClient/screens/apiClient/components/views/http/components/PathVariableTable/PathVariableTableRow.tsx
```diff
@@ -0,0 +1,94 @@
...
+  children,
+  dataIndex,
+  record,
+  environmentVariables,
+  handleUpdateVariable,
+  ...restProps
+}) => {
+  const form = useContext(EditableContext);
+  if (!form) {
+    Sentry.captureException(new Error("EditableCell rendered outside context"));
```
nsrCodes: renmae the exception to indicate it is pathVariables related

app/src/features/apiClient/screens/apiClient/components/views/http/components/PathVariableTable/PathVariableTableRow.tsx
nsrCodes: EditableRow infact is exactly the same, why are we making a copy of `EditableRow` when it already exists in KeyValueTableRow?

nsrCodes (changes requested):
tested, works well!

app/src/features/apiClient/screens/apiClient/components/views/http/components/PathVariableTable/PathVariableTableRow.tsx
RuntimeTerror10: The Path Variables table will not include checkbox UI elements. Since we have multiple editable table instances (including Environment tables), we'll need to create a base context and a reusable EditableRow component to standardize behavior across all instances.

For now, I've updated the naming conventions to reduce confusion.

app/src/features/apiClient/screens/apiClient/utils.ts
RuntimeTerror10: Updated regex. Now allows special characters which are not reserved in URLs

PR #3585, opened 2025-09-29
remove async iterator

nafees87n:
Async Iterator wasn't adding any value as requests are being iterated sequentially anyway


nafees87n:
Not relevant now. Closed

PR #3590, opened 2025-09-29
[ENGG-4461]: block navigation when collection running

rohanmathur91:


rohanmathur91:
Fixed in https://github.com/requestly/requestly/pull/3594

PR #3594, opened 2025-09-30
[ENGG-4461]: add isClosable tab types

rohanmathur91:


app/src/componentsV2/Tabs/store/tabStore.ts
```diff
@@ -10,25 +24,81 @@ export type TabState = {
   preview: boolean;
   title: string;
   icon: ReactNode;
+  closeBlockers: Map<CloseTopic, CloseBlocker>;
```
iostreamer-X: Let's keep this as `Map<CloseTopic, Map<string, CloseBlocker>>`. A topic can be used/registered by many 'things'. For example, we could add support for running two collections from the same tab. Which makes the tuple of `(topic, id)` the unique identifier and not just `topic`.

iostreamer-X (changes requested):

PR #3642, opened 2025-10-07
[ENGG-4496] feat: virtualization in collection runner result view

nafees87n:
Major challenged faced was in calculating the height. Antd takes some time to collapse/expand due to which virtualised list was calculating wrong heights if calculated immediately on click. 

## Solution
- Some delay was introduced in calculating the heights. Heights were calculated after transition animation was completed. But this introduced some jitter when collapsing/expanding
- To solve the jitter, we have made use of the fact that for a particular iteration the collapsed and expanded height would be same.
- As all the iterations are expanded by default, we cache the expanded height on ref load and this height is used to resize immediately on click. The collapsed height is hardcoded.

## Summary by CodeRabbit

* **New Features**
  * Virtualized test results for multi-iteration runs, improving performance and responsiveness.
  * Dynamic row heights with smoother expand/collapse transitions.
  * Tabs now free resources when inactive for better memory usage.

* **Style**
  * Updated to a full-height, flexible column layout across the results and history views.
  * Improved scrolling behavior and content sizing to prevent clipping.
  * Cleaner collapse styling with transparent backgrounds and reduced borders.

* **Refactor**
  * Reduced unnecessary re-renders and more accurate iteration counting.
  * Preserves existing behavior for single-iteration results.

### DIFF
--- a/app/src/features/apiClient/screens/apiClient/components/views/components/Collection/components/CollectionRunnerView/components/RunResultView/HistoryDrawer/historyDrawer.scss
+++ b/app/src/features/apiClient/screens/apiClient/components/views/components/Collection/components/CollectionRunnerView/components/RunResultView/HistoryDrawer/historyDrawer.scss
@@ -35,16 +35,31 @@
 
     .ant-drawer-body {
       padding: 0;
+      height: 100%;
+      display: flex;
+      flex-direction: column;
     }
   }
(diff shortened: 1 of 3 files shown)


nafees87n:
Performance not up to the mark. Not going with this. Closing

PR #3647, opened 2025-10-07
[ONCALL] fix: cannot read properties of null (reading 'id')

Aarushsr12:
Sentry Issue : [Link](https://requestly.sentry.io/issues/6836898825/events/127712c39ee44f4ba25266b4d71c5bfd/)


app/src/features/apiClient/screens/apiClient/utils.ts
```diff
@@ -407,11 +407,10 @@ export const createBlankApiRecord = (
     newRecord.collectionId = collectionId;
   }
 
-  const result =
-    recordType === RQAPI.RecordType.COLLECTION
-      ? apiClientRecordsRepository.createCollection(newRecord as RQAPI.CollectionRecord)
-      : apiClientRecordsRepository.createRecord(newRecord as RQAPI.ApiRecord);
-
+  //without await race condition
+  const result = await (recordType === RQAPI.RecordType.COLLECTION
```
rohanmathur91, line 411: How does it leads to race condition without await? and whats the race condition?

rohanmathur91 (changes requested):

PR #3668, opened 2025-10-10
Add support to inject scope for each execution of a request

iostreamer-X:


app/src/features/apiClient/helpers/variableResolver/variable-resolver.ts
```diff
@@ -79,9 +79,9 @@ export class VariableHolder {
   }
 }
 
-function getScopes(parents: string[], stores: AllApiClientStores): Scope[] {
-  const scopes: Scope[] = [];
-  let currentScopeLevel = 0;
+function getScopes(parents: string[], stores: AllApiClientStores, scopes: Scope[] = []): Scope[] {
```
iostreamer-X, line 82: This is the central solution to achieve per request injected scope. Here, rather than starting from scratch for our `scopes` container, we can continue with what we are given, as a starting point.

So if one want to inject something, one just needs to provide it in argument.

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestExecutor.ts
```diff
@@ -87,11 +88,12 @@ export class HttpRequestExecutor {
   async execute(
     recordId: string,
     entry: RQAPI.HttpApiEntry,
-    abortController?: AbortController
+    abortController?: AbortController,
+    scopes?: Scope[]
```
iostreamer-X, line 92: And this is where provide `scopes` that should be used for the execution of given request.

app/src/features/apiClient/store/variables/variables.store.ts
```diff
@@ -123,3 +150,47 @@ export const createVariablesStore = (props?: { variables: VariableValues }) => {
     },
   }));
 };
+
+export const createDummyVariableStore = (variables: Record<string, string | number | boolean>) => {
```
iostreamer-X, line 154: And to create a scope, we can use this helper which when given a simple record of variables, converts it into `StoreApi<VariableState>`, which in turn is the central part of what a `Scope` is.

app/src/features/apiClient/store/apiClientFilesStore.ts
```diff
@@ -99,6 +104,10 @@ const createApiClientFilesStore = (appMode: "desktop") => {
           return doesFileExist;
         },
 
+        // TODO: This needs to be changed such that it doesn't delete entries based on given records.
```
iostreamer-X, line 107: This is an important todo, without this, any changes done to the record will simply remove data added by us for collection runner.

app/src/features/apiClient/store/apiClientFilesStore.ts
```diff
@@ -4,6 +4,11 @@ import { RequestContentType, RQAPI } from "../types";
 import { generateKeyValuePairs, isHttpApiRecord } from "../screens/apiClient/utils";
 import { useShallow } from "zustand/shallow";
 
+export enum FileFeature {
```
iostreamer-X, line 7: We will re-use the existing store we have to manage files "uploaded". From here we get the basic functionalities of getting the size or existence of file, we wouldn't need to re-invent the wheel here.

PR #3785, opened 2025-10-31
[ENGG-4628] fix: asynchronous setup of variables from api client scripts

nafees87n:
## Issue

Our script handler waits until variable is set in the db. Thats why request is slower. A 466ms request takes 3.4s to actually complete.

image

## Solution

Ideally DB write should happen asynchronously and whoever needs the updated variables can use from local state. Script worker shouldn't await for any callback from UI.

Even if the worker is killed, the callback would run and complete as it runs in the context of UI and not worker

## Summary by CodeRabbit

* **New Features**
  * Script callbacks now receive structured snapshots; callbacks run only when state changes.

* **Refactor**
  * Centralized script execution around a shared execution context and workload manager; reruns use fresh isolated context.
  * Local state exposes full snapshot and mutation status; callback types allow sync or async handlers.
  * Pre-request flow now re-prepares request after mutations.

* **Bug Fixes**
  * Better runtime guards and user-facing error notifications on script failures.

* **Chores**
  * Corrected module import paths and public type exposures.

### DIFF
--- a/app/src/features/apiClient/screens/apiClient/utils.ts
+++ b/app/src/features/apiClient/screens/apiClient/utils.ts
@@ -861,7 +861,13 @@ export const extractPathVariablesFromUrl = (url: string) => {
   }
 
   const urlWithScheme = addUrlSchemeIfMissing(url);
-  const pathname = new URL(urlWithScheme).pathname;
+  let pathname: string = "";
+  try {
+    pathname = new URL(urlWithScheme).pathname;
+  } catch (error) {
+    Logger.log("Invalid URL while extracting path variables:", error);
+    return [];
+  }
 
   // Allow all characters except URL reserved characters: : / ? # [ ] @ ! $ & ' ( ) * + , ; =
   // Also exclude whitespace and control characters for practical reasons


app/src/features/apiClient/helpers/modules/scriptsV2/worker/script-internals/scriptExecutionWorker/scriptExecutionWorker.ts
```diff
@@ -39,12 +39,12 @@ export class ScriptExecutionWorker implements ScriptExecutionWorkerInterface {
     }
   }
 
-  private async syncLocalDump(callback: ScriptWorkloadCallback) {
```
iostreamer-X, line 42: Let's keep this async, but the caller would not wait for it. So that to wait or not is caller's decision and the underlying function support "awaiting".

app/src/features/apiClient/helpers/modules/scriptsV2/worker/script-internals/scriptExecutionWorker/scriptExecutionWorker.ts
```diff
@@ -39,12 +39,12 @@ export class ScriptExecutionWorker implements ScriptExecutionWorkerInterface {
...
+  private syncLocalDump(callback: ScriptWorkloadCallback) {
     if (!this.localScope.getIsStateMutated()) {
       return;
     }
     const dump = this.localScope.getAll();
-    await callback(dump);
+    callback(dump);
```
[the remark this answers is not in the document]
nafees87n: The errors are propagated to UI and handled there

iostreamer-X:
image

iostreamer-X:
Referring: https://github.com/requestly/requestly/pull/3785#issuecomment-3481093787

We can create a dedicated class for capturing the idea of a snapshot/execution context, rather than burdening scriptExecutor(existing implementation).

The idea is to think of this new class as storage, any state mutated by the script reflects on this storage.

And then when script execution completest(be it pre or post), the state of this storage provided to `postScriptExecutionCallback`.

This leads the setup where:
1. The executor builds the context(equivalent to `buildBaseSnapshot`)
2. Provides this context to `scriptExecutor`
3. `scriptExecutor`, rather than naively calling the `callback` provided, syncs the mutated state produced by worker to the context.
4. `httpRequestExecutor` then simply calls `postScriptExecutionCallback` with the current state of context


This keeps the actual syncing to database async, while still having a singular context for execution that lasts for pre and post scripts.

app/src/features/apiClient/helpers/httpRequestExecutor/scriptExecutionContext.ts
```diff
@@ -0,0 +1,95 @@
...
+    return this.context;
+  }
+
+  public getIsMutated(): boolean {
+    return this.isMutated;
+  }
+
+  public updateResponse(response: SnapshotForPostResponse["response"]) {
+    this.context.response = response;
+  }
```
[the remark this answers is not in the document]
nafees87n: That is intentional. Do not want the flag to be set when updating response

app/src/modules/localScope.ts
```diff
@@ -1,13 +1,17 @@
+import { ScriptExecutionContext } from "features/apiClient/helpers/httpRequestExecutor/scriptExecutionContext";
+
+export type InitialState = ScriptExecutionContext["context"];
```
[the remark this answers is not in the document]
nafees87n: Fixed in latest state

app/src/features/apiClient/helpers/modules/scriptsV2/worker/script-internals/scriptExecutionWorker/scriptExecutionWorker.ts
```diff
@@ -39,12 +39,12 @@ export class ScriptExecutionWorker implements ScriptExecutionWorkerInterface {
     }
   }
 
-  private async syncLocalDump(callback: ScriptWorkloadCallback) {
-    if (!this.localScope.getIsStateMutated()) {
-      return;
-    }
+  private syncLocalDump(callback: ScriptWorkloadCallback) {
```
iostreamer-X: Keep this promisified, we will just choose not to await this.

app/src/features/apiClient/helpers/httpRequestExecutor/scriptExecutionContext.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  public getContext(): ExecutionContext {
+    return this.context;
+  }
+
+  public getIsMutated(): boolean {
+    return this.isMutated;
+  }
+
+  public updateResponse(response: SnapshotForPostResponse["response"]) {
```
iostreamer-X: Let's rename this to `setResponse`

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestScriptExecutionService.ts
```diff
@@ -1,92 +1,38 @@
...
-    abortController: AbortController
-  ) {
+  async executePostResponseScript(entry: RQAPI.HttpApiEntry, abortController: AbortController) {
     return this.workloadManager.execute(
       new PostResponseScriptWorkload(
         entry.scripts?.postResponse,
-        this.buildPostResponseSnapshot(recordId, entry),
-        callback
+        this.executionContext.getContext(),
+        (snapshot: ScriptExecutionContext["context"]) => {
```
iostreamer-X: This is regarding the point you raised previously, where `ScriptExecutionContext` has another member inside called `context`. Let's keep that going. There's no snapshot anymore, only `context`

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestExecutor.ts
```diff
@@ -189,20 +194,18 @@ export class HttpRequestExecutor {
...
 
     if (
       preparedEntry.scripts?.preRequest.length &&
       preparedEntry.scripts?.preRequest !== DEFAULT_SCRIPT_VALUES[RQAPI.ScriptType.PRE_REQUEST]
     ) {
       trackScriptExecutionStarted(RQAPI.ScriptType.PRE_REQUEST);
-      preRequestScriptResult = await this.scriptExecutor.executePreRequestScript(
-        recordId,
-        preparedEntry,
-        this.postScriptExecutionCallback,
```
iostreamer-X, line 203: This call back used to be called immediately after script execution finished, regardless of it being pre or post.

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestExecutor.ts
```diff
@@ -289,13 +288,22 @@ export class HttpRequestExecutor {
       }
     }
 
+    const isSnapshotMutated = executionContext.getIsMutated();
+    if (isSnapshotMutated) {
+      this.postScriptExecutionCallback(executionContext.getContext());
```
iostreamer-X: Now, this is called only after pre AND post have completed. Let's call it like it used to before. We want to maintain the isolation that we had before. i.e Pre-Script is its own thing with its own lifecycle, same for post.

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestScriptExecutionService.ts
```diff
@@ -1,92 +1,48 @@
...
     return this.workloadManager.execute(
-      new PreRequestScriptWorkload(entry.scripts?.preRequest, this.buildPreRequestSnapshot(recordId, entry), callback),
+      new PreRequestScriptWorkload(
+        entry.scripts?.preRequest,
+        this.executionContext.getContext(),
+        (context: ScriptExecutionContext["context"]) => {
+          this.executionContext.updateContext(context);
+          postExecutionCallback(context);
+        }
+      ),
```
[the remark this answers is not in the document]
nafees87n: Fixed in commit [6f4954f](https://github.com/requestly/requestly/pull/3785/commits/6f4954f0d112d8f82ebf6e1d014b11b68f44bc67)

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestExecutor.ts
```diff
@@ -262,12 +270,13 @@ export class HttpRequestExecutor {
...
       trackScriptExecutionStarted(RQAPI.ScriptType.POST_RESPONSE);
-      responseScriptResult = await this.scriptExecutor.executePostResponseScript(
-        recordId,
-        preparedEntry,
-        this.postScriptExecutionCallback,
-        this.abortController
-      );
+      executionContext.setResponse(preparedEntry.response);
+      responseScriptResult = await scriptExecutor.executePostResponseScript(preparedEntry, this.abortController, () => {
+        const isSnapshotMutated = executionContext.getIsMutated();
```
iostreamer-X, line 278: This would return true regardless, if it was mutated in pre-script?

iostreamer-X (changes requested):

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestExecutor.ts
nafees87n: Yes

Issue #1124, opened 2023-10-06
bug: Requestly Firefox Devtools title shows both emoji and icon


wrongsahil:
### Describe your issue?

Requestly Firefox Devtools header is showing both emoji and png icon. This is okay in chromium based browsers as it shows icon only

### Repro steps

NA

### What Requestly tool were you using?

- [X] Extension
- [ ] Desktop-App
- [ ] Android SDK
- [ ] Selenium

### Your Environment

MacOs Ventura 13.0.1, Firefox

### Requestly Version

v23.9.20

### Error screenshot

<img width="249" alt="image" src="https://github.com/requestly/requestly/assets/16779465/c1348032-3dfe-4c68-9e18-a2d0dc9f1ccd">

adarsh-sgh:
> Requestly Firefox Devtools header is showing both emoji and png icon. This is okay in chromium based browsers as it shows icon only

Regarding the Chromium behavior, it displays the emoji only in my case. This makes sense since the icon feature was removed, as indicated in this conversation on [Superuser ](https://superuser.com/a/660003/1856667) .

Here is the code responsible for the current behavior, 

https://github.com/requestly/requestly/blob/e66ab075d880caa8a8681836795b94f405153fe8/browser-extension/common/src/devtools/devtools.js#L3

*** 

**Possible Solutions:**

**1. Conditionally render the emoji in Chrome only (Chrome : Emoji, Firefox : Icon):**
   - Chrome's behavior will remain unchanged.
   - Firefox will show the icon only, ensuring consistency with other panels in Firefox.
   - I am assuming we have some env variables distinguishing chrome builds from firefox ?

**2. Remove the emoji altogether:(Chrome : Text Only, Firefox : Icon)**
   - This would make it consistent with other panel items in Chrome .
   - Chrome will show just panel name, aligning with the behavior of other panels.
   - Firefox will show the icon, maintaining consistency with other panels.

SanchitMahajan236:
I want to contribute to this issue but can you please elaborate what approach is to be used to solve this issue as mentioned by adarsh

wrongsahil:
@SanchitMahajan236 If possible, please remove the icon from Firefox while keeping the emoji for both Chrome and Firefox. That would be ideal.

rasesh-here:
-- assign

SanchitMahajan236:
@wrongsahil please assign me this issue

ayaan-qadri:
Working on this !

Edit: Not working

[...]

PR #3790, opened 2025-11-02
fix: firefox dev tools icon

mustafa-sayyed:
Closes issue: #1124

## 📜 Summary of changes:
- Firefox was displaying icon and emoji both, even though we don't pass the icon path, it uses the default icon provided in the manifest file
- I made icon of that emoji to display it in Firefox
- if the Browser is Firefox, then it only shows text with emoji's icon
- Verified the behaviour by building extension and then tested it in FireFox

## 🎥 Demo Video:
Screenshot 2025-11-02 012800

## ✅ Checklist:

- [x] Verified UI in browser.
- [x] **Added demo video showing the changes in action** (if applicable).

## 🧪 Test instructions:
- Open the Firefox Devtools 

## Summary by CodeRabbit

* **New Features**
  * DevTools panel now adapts its title and icon based on browser detection — Firefox users see a Firefox-specific title and icon; other browsers see an adjusted title and no icon for a consistent experience.
  * Panel creation is now gated to active DevTools contexts so it appears only when an appropriate tab is open.


nafees87n (changes requested):
> Firefox was displaying icon and emoji both, even though we don't pass the icon path, it uses the default icon provided in the manifest file

If thats the case then can we display Requestly's icon in firefox and the emoji in chrome? We wanted to have consistency across but when we have to choose between a custom icon and Requestly logo in firefox then choosing the logo seems ideal rather than introducing a custom `png` icon.

mustafa-sayyed:
@nafees87n 
The Discussion happens in the issue 
@wrongsahil said to display emoji to maintain consistency

image

nafees87n:
@mustafa-sayyed 
I am aware about the discussion. That would have been ideal but thats not the case as you mentioned firefox always picks up the logo from manifest and to remove it you need to pass some other icon

> > Firefox was displaying icon and emoji both, even though we don't pass the icon path, it uses the default icon provided in the manifest file
> 
> If thats the case then can we display Requestly's icon in firefox and the emoji in chrome? We wanted to have consistency across but when we have to choose between a custom icon and Requestly logo in firefox then choosing the logo seems ideal rather than introducing a custom `png` icon.

This is the reason why we want to go down this route

mustafa-sayyed:
Ok, @nafees87n 

I will change to display the Requestly logo in Firefox instead of that emoji's icon

PR #4060, opened 2025-12-19
[ENGG-4739] feat: adds support of composite variables

nafees87n:
This adds support for composite variables. It enables using a variable as value for another variable.

Caveats:
1. Highlighting and autocompletion does not work in variables table.
2. If circular dependencies are detected then variables are left as is in unresolved form,

## Screenshots
Snipaste_2025-12-19_11-03-57
Snipaste_2025-12-19_11-03-23
Snipaste_2025-12-19_11-02-58
Snipaste_2025-12-19_11-02-55
Snipaste_2025-12-19_11-02-52

## Summary by CodeRabbit

* **New Features**
  * Variables can reference other variables; references are resolved in dependency order with handling for circular references and unresolved cases; failures fall back to original values.

* **Refactor**
  * Improved lifecycle and null-safety for the variable resolution manager to ensure reliable initialization and guarded usage.

* **Chores**
  * Added a new runtime dependency to support dependency-based variable resolution.


app/src/backend/environment/utils.ts
```diff
@@ -160,6 +164,77 @@ const collectAndEscapeVariablesFromTemplate = (
...
+      }
+    }
+  });
+
+  // Get resolution order (topologically sorted)
+  let resolutionOrder: string[];
+  try {
+    resolutionOrder = graph.overallOrder();
+  } catch (error) {
+    Logger.error("Failed to resolve variable dependencies:", error);
```
rohanmathur91: Should we capture this on sentry?

app/src/backend/environment/utils.ts
```diff
@@ -160,6 +164,77 @@ const collectAndEscapeVariablesFromTemplate = (
...
+          }
+        });
+      }
+    }
+  });
+
+  // Get resolution order (topologically sorted)
+  let resolutionOrder: string[];
+  try {
+    resolutionOrder = graph.overallOrder();
```
rohanmathur91: Why will this fail? due to cycle in graph?

PR #4079, opened 2025-12-24
feat: ENGG-5124 - Detection of non-supported features in Postman Importers

rdsharma3333:
Closes issue: ENGG-5124

## 📜 Summary of changes:

Fixed two things:

1. Silent Import Failures : When users imported Postman collections with unsupported authentication types (like OAuth2, AWS Auth, Digest Auth), the import would silently fail and nothing would appear in the API Client. Users had no idea what went wrong.
2.Detection of non-supported features in Postman Importers and then firing the event for analytics

## 🎥 Demo Video:

**Video/Demo:** 

## ✅ Checklist:

- [ ] Verified UI in browser.

## Summary by CodeRabbit

* **New Features**
  * Postman importer now detects and aggregates unsupported Postman features (vault/dynamic/disabled vars, unsupported auth/scripts, request examples) with richer per-collection and per-environment metadata.
  * Import emits a telemetry event when unsupported features are detected for improved import visibility and reporting.

* **Bug Fixes**
  * Improved error handling and resilience so individual item errors no longer break overall import processing.


wrongsahil (changes requested):
.

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/PostmanImporter.tsx
```diff
@@ -116,16 +128,81 @@ export const PostmanImporter: React.FC<PostmanImporterProps> = ({ onSuccess }) =
             apiRecords: [],
             variables: {},
           };
+          const allUnsupportedFeatures = new Set<string>();
+          const allMeta: Partial<UnsupportedFeaturesMeta> = {
```
nafees87n: We shouldn't use Partial here

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/utils.ts
```diff
@@ -276,44 +611,74 @@ const createApiRecord = (
...
+          body: "",
+          bodyContainer: createBodyContainer({ contentType: RequestContentType.RAW, body: "" }),
+          contentType: RequestContentType.RAW,
+        },
+        response: null,
+        auth: getDefaultAuth(parentCollectionId === null),
+        scripts: { preRequest: "", postResponse: "" },
+      },
+    } as Partial<RQAPI.HttpApiRecord>;
+  }
```
nafees87n, line 681: Should we continue importing using defaults? Shouldn't we throw error here?
Users might get confused because something else was imported which is unintended 
@wrongsahil @rdsharma3333

rdsharma3333: Previously we were failing the import silently. My thought was we should not fail, instead we should still import the collection with whatever gets imported. Today we might not be supporting few func but later we will be able to make it work by default.

nafees87n: But in that particular part of code, it would not fail because of unsupported feature but faulty code. All the features have been handled to have a default value if something unsupported comes up.
So silently failing there would seem inconsistent

PR #4080, opened 2025-12-26
[PLAT-17] ref: Migrate Environment view to use Redux buffer

srbh777:
Closes issue: [PLAT-17](https://linear.app/requestly/issue/PLAT-17)

📜 **Summary of changes:**
This pull request refactors the `EnvironmentView` component, migrating its state management from the existing Zustand-based implementation to the new Redux-based buffer architecture. This change aligns environment handling with the new standard for managing API client entities, providing a consistent draft-and-save workflow.

The core of this refactor involves introducing buffered entities for environments, which allows changes to be tracked in a draft state before being explicitly saved. Persistence is now handled by dedicated Redux thunks, which communicate with the repository and update the central state upon success.

🎥 **Demo Video:**
Video/Demo: *(Add link or upload demo)*

✅ **Checklist:**
- [ ] Linting and unit tests pass
- [ ] No install/build warnings introduced
- [ ] UI verified in browser
- [ ] Analytics updated (if applicable)
- [ ] Extension tested in Chrome & Firefox (if applicable)
- [ ] Unit tests added/updated
- [ ] Docs updated (if applicable)
- [ ] Demo video added (if applicable)

🧪 **Test instructions:**
- *(Add clear test steps here)*

🔗 **Other references:**
- **`EnvironmentView` Component:**
  - Replaced Zustand hooks (`useCommand`, `useEnvironment`) with Redux hooks (`useBufferedEnvironmentEntity`, `useApiClientDispatch`).
  - Integrated the component with the Redux buffer system to manage draft states for environment variables.
  - On mount, the component now opens a buffer for the current environment and closes it on unmount.
  - Saving changes now marks the buffer as saved, decoupling the UI from direct persistence logic.
  - Unsaved changes are now tracked using the buffer's `isDirty` state.

- **Redux Slices & Entities:**
  - Created `BufferedEnvironmentEntity` and `BufferedGlobalEnvironmentEntity` to represent environments within the buffer.
  - Updated the `BufferedEntityFactory` to construct these new entity types.
  - Added `updateEnvironmentVariables` and `updateEnvironmentName` async thunks for persisting changes.
  - Configured the `environments` slice to listen for `entitySynced` actions, allowing it to update after successful persistence.

- **Redux Buffer System:**
  - Extended the buffer middleware to monitor the `environments` slice for external changes.
  - Added `ENVIRONMENT` and `GLOBAL_ENVIRONMENT` types to the buffer's entity map and source selectors.

- **Hooks & Types:**
  - Introduced `useBufferedEnvironmentEntity` hook for convenient access to buffered environment instances.
  - Updated `useApiClientDispatch` to support `ThunkDispatch` for async actions.
  - Added `EnvironmentsState` to the `ApiClientRootState` type definition.

### DIFF
--- a/app/src/features/apiClient/slices/buffer/index.ts
+++ b/app/src/features/apiClient/slices/buffer/index.ts
@@ -1,5 +1,5 @@
 export { bufferSlice, bufferActions, bufferReducer, bufferAdapterSelectors } from "./slice";
 export { bufferSyncMiddleware, bufferListenerMiddleware } from "./middleware";
 export { saveBuffer } from "./thunks";
-export type { BufferEntry, NewBufferEntry, EditBufferEntry, BufferState, EntityDataMap } from "./types";
+export type { BufferEntry, NewBufferEntry, EditBufferEntry, BufferState } from "./types";
 export * from "./hooks";
--- a/app/src/features/apiClient/slices/buffer/middleware.ts
+++ b/app/src/features/apiClient/slices/buffer/middleware.ts
@@ -37,10 +37,10 @@ const BUFFER_SOURCE_SELECTORS: SourceSelectorMap = {
     sliceName: API_CLIENT_ENVIRONMENTS_SLICE_NAME,
   },
   [ApiClientEntityType.GLOBAL_ENVIRONMENT]: {
-    selectById: (state, id) =>
-      state.environments.globalEnvironment.id === id ? state.environments.globalEnvironment : undefined,
+    selectById: (state) =>
+      state.environments.globalEnvironment,
     sliceName: API_CLIENT_ENVIRONMENTS_SLICE_NAME,
-  },
+  },    
 };
 
 const MONITORED_ROOT_SLICES: readonly string[] = [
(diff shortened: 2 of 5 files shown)


app/src/features/apiClient/screens/environment/components/environmentView/EnvironmentView.tsx
```diff
@@ -9,42 +10,96 @@ import { ApiClientExportModal } from "features/apiClient/screens/apiClient/compo
...
-    return mapToEnvironmentArray(Object.fromEntries(variablesMap.data));
-  }, [variablesMap]);
-  const {
-    env: { setEnvironmentVariables },
-  } = useCommand();
+  const dispatch = useApiClientDispatch();
+  // const repositories = useApiClientRepository();
+
+  // Determine if this is global environment
+  const isGlobal = envId === GLOBAL_ENVIRONMENT_ID;
```
iostreamer-X: If an env is global or not, will come from the top as a prop, we will not checks it based on a magic string.
Secondly, we will change the `EnvironmentView` such that it doesn't figure out the entity based on id, but simply gets it from its props and goes about its way as it would, without worrying if it's global or non-global.

With that done, we now simply need a small component that will create the correct entity and pass it to `EnvironmentView`. This small router/manager component is what one would mount in-order to render the environment, and this is what would be open when clicking on a row to open a tab.

iostreamer-X (approved):

Issue #4026, opened 2025-12-12
Support Comments Inside Request Body (for API Testing)


chandanPradhan09:
**Description**
While testing APIs, it’s often useful to include comments within the request body—especially when working with large JSON payloads, optional fields, or conditional data. Currently, Requestly does not allow adding comments (e.g., // comment) inside the body of a request. This causes the request to fail because the body becomes invalid JSON.

**Why This Is Important**

Helps developers annotate complex payloads

Makes debugging easier

Useful for toggling optional fields during testing

Matches behavior of tools like Postman & Insomnia (which gracefully ignore comments)

**Proposed Solution**
Allow comments in request bodies by stripping them before sending the final API request.
Support for formats like:

{
  "name": "Chandan",
  // "testValue": 123,  // temporarily disabled
  "active": true
}

**Additional Notes**

This would greatly enhance usability for API developers and testers.

Anusha-pannati:
heyy!!, I would like to solve this issue.

Kanishkrawatt:
Hey @chandanPradhan09, Support for comments in the request body has now been added!

PR #4110, opened 2026-01-05
[ENGG-4693] feat: Add comments support for JSON modes in API Client editors

Kanishkrawatt:
Closes issue: #4026
## 📜 Summary of changes:
Added support for comments in JSON request bodies for the API client

## 🎥 Demo Video:

https://github.com/user-attachments/assets/de7ae41d-15f2-4aa8-b7cc-772749a2cfbf

## ✅ Checklist:

- [x] Verified UI in browser.
- [x] For UI changes, added/updated analytics events (if applicable).
- [x] **Added demo video showing the changes in action** (if applicable).

## Summary by CodeRabbit

* **New Features**
  * Added JSON5 support in the code editor, enabling more flexible JSON syntax (comments, trailing commas) for editing and highlighting.
  * API client now normalizes and accepts JSON5-formatted request bodies, improving handling of relaxed JSON input.

### DIFF
--- a/app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
+++ b/app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
@@ -3,6 +3,7 @@ import CodeMirror, { EditorView, ReactCodeMirrorRef } from "@uiw/react-codemirro
 import { javascript } from "@codemirror/lang-javascript";
 import { html } from "@codemirror/lang-html";
 import { json5 } from "codemirror-json5";
+import { json } from "@codemirror/lang-json";
 import { css } from "@codemirror/lang-css";
 import { vscodeDark } from "@uiw/codemirror-theme-vscode";
 import { EditorLanguage, EditorCustomToolbar, AnalyticEventProperties } from "componentsV2/CodeEditor/types";
@@ -126,6 +127,8 @@ const Editor: React.FC<EditorProps> = ({
       case EditorLanguage.JAVASCRIPT:
         return javascript({ jsx: false });
       case EditorLanguage.JSON:
+        return json();
+      case EditorLanguage.JSON5:
         return json5();
       case EditorLanguage.HTML:
         return html();
--- a/app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestPreparationService.ts
+++ b/app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestPreparationService.ts
@@ -204,10 +204,11 @@ export class HttpRequestPreparationService {
       typeof renderedEntry.request.body === "string"
     ) {
       try {
+        // Here we are parsing and then re-stringifying to remove any comments and trailing commas from JSON
         const parsed = JSON5.parse(renderedEntry.request.body);
         renderedEntry.request.body = JSON.stringify(parsed);
       } catch (error) {
-        console.warn("Failed to parse JSON5 body:", error);
+        // No op - If parsing fails, we leave the body as it is
       }
     }
 
--- a/app/src/componentsV2/CodeEditor/types/index.ts
+++ b/app/src/componentsV2/CodeEditor/types/index.ts
@@ -5,6 +5,7 @@ import { ReactNode } from "react";
 export enum EditorLanguage {
   JAVASCRIPT = "javascript",
   JSON = "json",
+  JSON5 = "json5",
   HTML = "html",
   CSS = "css",
 }
--- a/app/src/features/apiClient/screens/apiClient/components/views/components/request/renderers/raw-body-renderer.tsx
+++ b/app/src/features/apiClient/screens/apiClient/components/views/components/request/renderers/raw-body-renderer.tsx
@@ -28,7 +28,7 @@ export function RawBody(props: {
     { leading: true, trailing: true }
   );
 
-  const editorLanguage = contentType === "application/json" ? EditorLanguage.JSON : null;
+  const editorLanguage = contentType === "application/json" ? EditorLanguage.JSON5 : null;
 
   return (
     <div className="api-client-code-editor-container api-request-body-editor-container">


app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestPreparationService.ts
```diff
@@ -196,6 +198,18 @@ export class HttpRequestPreparationService {
...
     );
     renderedEntry.request.url = addUrlSchemeIfMissing(renderedEntry.request.url);
+    if (
+      renderedEntry.request.body &&
+      renderedEntry.request.contentType === RequestContentType.JSON &&
+      typeof renderedEntry.request.body === "string"
+    ) {
+      try {
+        const parsed = JSON5.parse(renderedEntry.request.body);
+        renderedEntry.request.body = JSON.stringify(parsed);
```
RuntimeTerror10, line 209: Add comment here why an extra step for parsing the stringified JSON is required.

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestPreparationService.ts
```diff
@@ -196,6 +198,18 @@ export class HttpRequestPreparationService {
       renderedEntry.request.pathVariables || []
     );
     renderedEntry.request.url = addUrlSchemeIfMissing(renderedEntry.request.url);
+    if (
+      renderedEntry.request.body &&
+      renderedEntry.request.contentType === RequestContentType.JSON &&
+      typeof renderedEntry.request.body === "string"
```
RuntimeTerror10, line 204: No need to check for `string` type

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestPreparationService.ts
```diff
@@ -196,6 +198,18 @@ export class HttpRequestPreparationService {
...
+    ) {
+      try {
+        const parsed = JSON5.parse(renderedEntry.request.body);
+        renderedEntry.request.body = JSON.stringify(parsed);
+      } catch (error) {
+        console.warn("Failed to parse JSON5 body:", error);
+      }
```
RuntimeTerror10, line 212: Add a no op here since malformed JSON is already present in the body

app/src/componentsV2/CodeEditor/components/EditorV2/Editor.tsx
```diff
@@ -126,7 +126,7 @@ const Editor: React.FC<EditorProps> = ({
       case EditorLanguage.JAVASCRIPT:
         return javascript({ jsx: false });
       case EditorLanguage.JSON:
-        return json();
+        return json5();
```
RuntimeTerror10, line 132: Create a new enum for JSON5 and use this for all editors in API Client

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestPreparationService.ts
Kanishkrawatt: We cannot remove the string check because TypeScript throws an error saying the request body can be FormData or multipart form.

Even though we are checking the content type to confirm it is JSON, TypeScript cannot infer from that check that the request body will always be a string when the content type is JSON.

app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestPreparationService.ts
```diff
@@ -196,6 +198,18 @@ export class HttpRequestPreparationService {
...
+    }
```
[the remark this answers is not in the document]
Kanishkrawatt: This is expected behaviour. If JSON5 parsing fails, it means the input JSON is invalid. In that case, we intentionally do not modify or block the request and allow it to be sent as is. The server is responsible for handling and responding to malformed payloads, and this keeps the client behaviour non intrusive and predictable.

Kanishkrawatt:
## Why choosing [JSON5](https://json5.org/) and not [JSONC](https://jsonc.org/) ?

First, there is an CodeMirror language package available for [JSON5](https://github.com/dimfeld/codemirror-json5), but not for JSONC. This makes editor integration much more easier.

Second, JSON5 supports features like trailing commas. In JSONC, commenting out a field can easily introduce a trailing comma, which results in invalid JSON. Since we do not have editor Linting at the moment, users would not get any immediate feedback about this issue and would end up sending a malformed request that returns a 400 Bad Request.

The main use case for allowing comments in JSON is to temporarily disable a specific field for a particular request.

For example:

```json
{
  "bar": "foo",
  "bar2": "foo2"
}
```

If a user comments out `bar2`:

```json
{
  "bar": "foo", ---> This comma 
  //"bar2": "foo2"
}
```

This introduces a trailing comma, which is invalid in standard JSON and JSONC. Without Linting, this becomes a poor user experience.

JSON5 solves this problem. When we parse the body using the JSON5 parser, it safely handles comments and removes trailing commas, allowing the request to go through successfully without forcing users to manually fix formatting issues.

Therefore, JSON5 makes more sense for this.

wrongsahil:
> ## Why choosing [JSON5](https://json5.org/) and not [JSONC](https://jsonc.org/) ?
> 
> 
> 
> First, there is an CodeMirror language package available for [JSON5](https://github.com/dimfeld/codemirror-json5), but not for JSONC. This makes editor integration much more easier.
> 
> 
> 
> Second, JSON5 supports features like trailing commas. In JSONC, commenting out a field can easily introduce a trailing comma, which results in invalid JSON. Since we do not have editor Linting at the moment, users would not get any immediate feedback about this issue and would end up sending a malformed request that returns a 400 Bad Request.
> 
> 
> 
> The main use case for allowing comments in JSON is to temporarily disable a specific field for a particular request.
> 
> 
> 
> For example:
> 
> 
> 
> ```json
> 
> {
> 
>   "bar": "foo",
> 
>   "bar2": "foo2"
> 
> }
> 
> ```
> 
> 
> 
> If a user comments out `bar2`:
> 
> 
> 
> ```json
> 
> {
> 
>   "bar": "foo", ---> This comma 
> 
>   //"bar2": "foo2"
> 
> }
> 
> ```
> 
> 
> 
> This introduces a trailing comma, which is invalid in standard JSON and JSONC. Without Linting, this becomes a poor user experience.
> 
> 
> 
> JSON5 solves this problem. When we parse the body using the JSON5 parser, it safely handles comments and removes trailing commas, allowing the request to go through successfully without forcing users to manually fix formatting issues.
> 
> 
> 
> Therefore, JSON5 makes more sense for this.

Looks good. 
Point to note: We should always ensure that the final output from the editor is always parsed and converted into a valid json object first. This would always ensure that served receives a valid hson object else it'd fail

RuntimeTerror10:
@wrongsahil I think we should not give editor the responsibility of always returning valid JSON. Editor should be a simple textarea with additional formatting capabilities and should not restrict the user the type of data they get from editor changes. 

For example: In API Client , If I enter an invalid JSON I will expect the server to give me a `bad request` as a response and here editor should not barge in to make it valid.

Also if we editor parses the JSON then source (state) will not have comments in the code, which is a bug.

We can give a feedback for invalid JSON while user is typing/pasting by using CodeMirror's linter plugin. This can give user a  visual error lines so that user can see it before taking any further action.

If we treat the editor as a plain-text source of truth with a linter, we preserve the user's formatting and comments

PR #4205, opened 2026-01-21
[ENGG-5195] fix: remove noise from sentry new dashboards

Aarushsr12:
* **Improvements**
  * Expanded error capture across sync and local store operations to surface individual failures for better monitoring.
  * Added targeted error reporting and tagging during Postman import flows to improve detection of partial or failed imports.
  * Hardened handling of authentication options (Bearer, Basic, API key) to avoid runtime errors and missing values.
* **Behavior**
  * Import flows now propagate and report failures more consistently.

This PR also fixes, these sentry errors: 

1. https://requestly.sentry.io/issues/6569685647/?project=4503895961305088
2. https://requestly.sentry.io/issues/6540669618/?project=4503895961305088
3. https://requestly.sentry.io/issues/6499659276/?project=4503895961305088


app/src/features/apiClient/helpers/modules/sync/localStore/services/LocalStoreRecordsSync.ts
```diff
@@ -105,8 +106,13 @@ export class LocalStoreRecordsSync implements ApiClientRecordsInterface<ApiClien
...
 
-    await this.queryService.createRecord(newRecord);
-    return { success: true, data: newRecord };
+    try {
+      await this.queryService.createRecord(newRecord);
+      return { success: true, data: newRecord };
+    } catch (error) {
+      captureException(error);
+      return { success: false, data: null };
+    }
```
nsrCodes:
```suggestion
        return this.queryService.createRecord(newRecord)
      .then(() => {
        return { success: true as true , data: newRecord };
      }).catch((error) => {
        captureException(error);
        return { success: false, data: null, message: error.message };
      })
```

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/PostmanImporter.tsx
```diff
@@ -230,6 +236,8 @@ export const PostmanImporter: React.FC<PostmanImporterProps> = ({ onSuccess }) =
       collections,
       handleCollectionWrites
     );
+
+    // should we cleanup such errors, this will also help in cleanup of sentry?
```
nsrCodes: not right now. these are product errors and we'll resolve them later once we are sure sentry is now setup properly with the new errors

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/PostmanImporter.tsx
```diff
@@ -291,6 +299,23 @@ export const PostmanImporter: React.FC<PostmanImporterProps> = ({ onSuccess }) =
...
+              Sentry.withScope((scope) => {
+                scope.setTag("error_type", "api_client_postman_import_environments_failed");
+                Sentry.captureException(environmentsResult.reason);
+              });
+            }
+
+            if (collectionsResult.status === "rejected") {
+              Sentry.withScope((scope) => {
+                scope.setTag("error_type", "api_client_postman_import_collections_failed");
+                Sentry.captureException(collectionsResult.reason);
```
nsrCodes: can we trigger this from local and confirm the traces are coming in with all the correct details?

app/src/features/apiClient/helpers/modules/sync/cloud/services/FirebaseApiClientRecordsSync.ts
```diff
@@ -182,13 +183,21 @@ export class FirebaseApiClientRecordsSync implements ApiClientRecordsInterface<A
...
   ) {
     try {
       const result = await batchWrite(batchSize, entities, writeFunction);
+      result.forEach((r) => {
+        if (r.status === "rejected") {
+          const err = r.reason;
+          captureException(err);
+        }
+      });
+
```
nsrCodes, line 200: have we checked this isn't causing perf issues on first load or bulk imports?

Aarushsr12: Nope this is not causing any performance issues on first load & while testing the bulk imports having 150 API requests

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/PostmanImporter.tsx
Aarushsr12: I tried to do this, but was not able to bring the rejected state, hence was not able to trigger this

Aarushsr12: Also I have removed the environmentsResult rejected `if` as we are already capturing the result reason inside the  `handleImportEnvironments`

Aarushsr12: this for `handleImportCollectionsAndApis`
image
image

app/src/features/apiClient/helpers/modules/sync/cloud/services/FirebaseApiClientRecordsSync.ts
Aarushsr12: If processing is successful but breaks at import, to write a collection it uses `batchwriteApiEntities` function
hence the above throws such error on sentry
image

image

Aarushsr12:
Inside onSave function it makes call to `createRecordwithId` function which internally calls the createApiRecord which holds the sentry checks, verified by logging this error
image

app/src/features/apiClient/screens/PostmanImporterView/components/PostmanImporter/PostmanImporter.tsx
```diff
@@ -216,12 +222,11 @@ export const PostmanImporter: React.FC<PostmanImporterProps> = ({ onSuccess }) =
...
         } else {
           failedCollectionsCount++;
-          return null;
         }
       } catch (error) {
         failedCollectionsCount++;
         Logger.error("Error importing collection:", error);
-        return null;
+        throw error;
       }
```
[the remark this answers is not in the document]
Aarushsr12: yes, this will break on the very first collection that is failure and interrupts the batch processing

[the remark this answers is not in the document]
Aarushsr12: but this also leads to not throw error for those cases where collectionResult is marked fulfilled but import failed, instead it should be rejected state, that why I added throw error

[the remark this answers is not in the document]
Aarushsr12: This will come more under product error rather than system error

Issue #4197, opened 2026-01-20
HTTP call using pre-request script


rahulGarg003:
Add a feature where user can fetch a auth token using a http api call before calling their exact api

rohanmathur91:
Hi @rahulGarg003, we have released this in https://github.com/requestly/requestly/pull/4215.
Refer below for the same. 


Note: Currently we only support `async & await` syntax.

```ts
// pre-request script

const response = await fetch('https://httpbin.org/uuid') 
const token = await response.json()
rq.environment.set("uuid", token.uuid); // consume this in header
```

PR #4215, opened 2026-01-22
[ENGG-5192] feat: async pre and post scripts support

rohanmathur91:
# Demo:

https://github.com/user-attachments/assets/a50618e5-e9e1-4a29-9408-b09b9212d43f

# Testcases:
1. Promise based + sync setter methods
```ts
const p = new Promise(resolve => {
  setTimeout(() => resolve("done"), 3000) // Request should wait for atleast 3 secs
})


await p.then(async () => {
rq.environment.set("x-from-pre-script-before", "before fetch"); // consume this in header
  
const response = await fetch('https://app.requestly.io/echo') 

const r = await response.json()
const s = JSON.stringify(r)

rq.environment.set("x-from-pre-script-after", s); // consume this in header
})
```

2. Hmac
```ts
async function hmacWeb(secret, message) {
  const enc = new TextEncoder();

  // Convert to bytes
  const keyData = enc.encode(secret);
  const msgData = enc.encode(message);

  // Import key
  const cryptoKey = await crypto.subtle.importKey(
    "raw",
    keyData,
    { name: "HMAC", hash: { name: "SHA-256" } },
    false,
    ["sign"]
  );

  // Create HMAC
  const sig = await crypto.subtle.sign("HMAC", cryptoKey, msgData);

  // Convert ArrayBuffer → hex
  return [...new Uint8Array(sig)]
    .map(x => x.toString(16).padStart(2, "0"))
    .join("");
}


const code = await hmacWeb("my-secret", "hello world");

rq.test("hmac will not be equal", function() {
  rq.expect(code).to.equal(rq.environment.get("hmacWeb"));
});

rq.environment.set("hmacWeb", code); // consume this in header

rq.test("hmac will be equal", function() {
  rq.expect(code).to.equal(rq.environment.get("hmacWeb"));
});
```

3. Promise with `.then` chain (Won't work, failing case)
```ts
// NOTE: below will not work, script will NOT WAIT for async code to be resolved
fetch('https://app.requestly.io/echo')
.then(response => response.json())
.then(r => JSON.stringify(r))
.then(token => rq.environment.set("x-from-pre-script-after", token))
```

4. Promise methods
```ts
const [user, config] = await Promise.all([
  fetch('https://httpbin.org/json').then(r => r.json()),
  fetch('https://httpbin.org/uuid').then(r => r.json())
]);

rq.environment.set('user', user.slideshow.author);
rq.environment.set('config', config.uuid);
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved script execution to properly handle asynchronous operations and return results correctly.
  * Enhanced error handling to provide more reliable error reporting during script execution.

### DIFF
--- a/app/src/features/apiClient/helpers/modules/scriptsV2/worker/script-internals/scriptExecutionWorker/scriptExecutionWorker.ts
+++ b/app/src/features/apiClient/helpers/modules/scriptsV2/worker/script-internals/scriptExecutionWorker/scriptExecutionWorker.ts
@@ -77,7 +77,7 @@ export class ScriptExecutionWorker implements ScriptExecutionWorkerInterface {
     try {
       await scriptFunction(globalObject);
     } catch (error) {
-      throw new ScriptExecutionError(error);
+      throw new ScriptExecutionError(typeof error === "string" ? new Error(error) : error);
     }
     try {
       this.syncLocalDump(callback);


rohanmathur91:
Fixes: https://github.com/requestly/requestly/issues/4197

PR #4223, opened 2026-01-23
PLAT-66 Fix collection variables scoping in request preparation service

srbh777:
Closes issue: 

## 🎥 Demo Video:

**Video/Demo:** 

## ✅ Checklist:

- [ ] Verified UI in browser.


app/src/features/apiClient/helpers/httpRequestExecutor/httpRequestPreparationService.ts
```diff
@@ -43,14 +43,27 @@ export class HttpRequestPreparationService {
...
+    const parents = selectAncestorIds(state, recordId);
+
+    // Build collection variables map - apply execution context's collectionVariables to all collection IDs in hierarchy
+    const collectionVariablesMap: Record<string, typeof context.collectionVariables> = {};
+    if (context.collectionVariables && Object.keys(context.collectionVariables).length > 0) {
+      // Apply to the recordId if it's a collection, and all parent collections
+      for (const id of [recordId, ...parents]) {
+        const record = selectRecordById(state, id);
+        if (record?.type === RQAPI.RecordType.COLLECTION) {
+          collectionVariablesMap[id] = context.collectionVariables;
```
iostreamer-X, line 57: We have received `collectionVariables`, which is a simple `{[key]: VariableData}` and then seemingly creating a map where for every collection in our ancestry we set `collectionVariables` as the corresponding variable that exist for that collection.

Firstly, our input is `collectionVariables` which doesn't tell which collection these variables belong to, and our output is variables for every collection in ancestry. We just extrapolated the input into fitting the output we need.

- How does this solve PLAT-66?
- If it's solving it, we need to dig deeper? Why an extrapolated override solved this bug about set collection variables from script?

iostreamer-X (changes requested):

PR #4326, opened 2026-02-11
[ENGG-5246] feat: add dynamic variable support in api client pre and post response script

rohanmathur91:
Demo:

https://github.com/user-attachments/assets/7975f7c4-8d77-47b4-9124-c996feee7914

### DIFF
--- a/app/src/lib/dynamic-variables/index.ts
+++ b/app/src/lib/dynamic-variables/index.ts
@@ -2,4 +2,4 @@ import { FakerVariableProvider } from "./providers/faker/FakerVariableProvider";
 import { HandlebarsResolver } from "./resolvers/handlebars";
 
 const fakerProvider = new FakerVariableProvider();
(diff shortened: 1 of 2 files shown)


app/src/lib/dynamic-variables/index.ts
```diff
@@ -5,4 +5,4 @@ export { DynamicVariableResolver } from "./resolvers/DynamicVariableResolver";
 export { HandlebarsResolver } from "./resolvers/handlebars";
 
 const fakerProvider = new FakerVariableProvider();
-export const dynamicVariableResolver = new HandlebarsResolver(fakerProvider);
+export const dynamicVariables = new HandlebarsResolver(fakerProvider);
```
iostreamer-X: Questionable. Can a resolver be called straight up dynamic variables? One would assume that dynamicVariables is `DynamicVariables[]`.

PR #4392, opened 2026-02-19
[ENGG-5335] fix: import graphQL requests in its view

Aarushsr12:
## Summary by CodeRabbit

* **New Features**
  * Added GraphQL import support for Postman collections, extracting operation, variables, and operation name.
  * GraphQL imports are now routed through a dedicated GraphQL import path to preserve headers and authentication during import.
  * Improved GraphQL body parsing for more reliable imports.

* **Bug Fixes**
  * Safer URL handling: missing or non-string URLs now default to an empty string to avoid import errors.

PR #4400, opened 2026-02-20
[ENGG-5335] fix: added fallback for operationName undefined in loggedIn state

Aarushsr12:
What the change does
- Before: The request object always had an operationName property, even when there was no name:
`operationName: graphqlBody.operationName` -> could be `undefined` or `""`.
undefined values leads to break while saving in firebase

- After: The property is only added when there is a real value
`...(operationName != null && operationName !== "" && { operationName })` → the key is omitted when it’s null, undefined, or empty string


Aarushsr12:
This is fix for graphql imports not coming in loggedIn state (https://github.com/requestly/requestly/pull/4392)

PR #4428, opened 2026-02-25
[DB-34] feat: secrets variable resolver

nafees87n:
1. Resolves secret variables using handlebar resolver
2. Uses a dummy source for now

## Release Notes

* **New Features**
  * Added support for secret variables as a new variable scope alongside existing scopes
  * Secrets are now integrated into template rendering and variable resolution
  * Secret variables are visually identified with a lock icon in the variable editor
  * Secret values are protected and cannot be modified through the editing interface


nafees87n:
@coderabbitai review

PR #4472, opened 2026-03-04
[ONCALL-336] fix: fallback for non-string values in queryParams

Aarushsr12:
## Summary by CodeRabbit

* **Bug Fixes**
  * Normalized non-string parameter and query values across import and API utilities: numbers and booleans are now coerced to strings; null/unsupported object values default to empty strings, ensuring consistent key/value generation and preventing data-type issues during imports.

### DIFF
--- a/app/src/features/apiClient/screens/apiClient/components/modals/importModal/utils.ts
+++ b/app/src/features/apiClient/screens/apiClient/components/modals/importModal/utils.ts
@@ -93,12 +93,7 @@ export const processRqImportData = (
         (queryParam: KeyValuePair, index: number) => {
           return {
             ...queryParam,
-            value:
-              typeof queryParam.value === "string"
-                ? queryParam.value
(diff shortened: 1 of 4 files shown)


app/src/features/apiClient/screens/apiClient/components/modals/importModal/utils.ts
```diff
@@ -93,6 +93,12 @@ export const processRqImportData = (
         (queryParam: KeyValuePair, index: number) => {
           return {
             ...queryParam,
+            value:
+              typeof queryParam.value === "string"
+                ? queryParam.value
+                : typeof queryParam.value === "number" || typeof queryParam.value === "boolean"
+                ? String(queryParam.value)
+                : "",
```
RuntimeTerror10: Can't we default to `String(val)`instead of checking for boolean and numbers everywhere?

PR #4475, opened 2026-03-05
Fixed collection variable sync on cloud for empty row

rohitneharabrowserstack:
Closes issue: 

## 🎥 Demo Video:

**Video/Demo:** 

## ✅ Checklist:

- [ ] Verified UI in browser.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed saving of collection variables in API records: keys are trimmed, empty keys are excluded, local-only values are removed, and variable ordering is reconciled with the remaining keys. This prevents stale or empty entries from persisting and ensures consistent variable maps when updating records.

### DIFF
--- a/app/src/backend/apiClient/upsertApiRecord.ts
+++ b/app/src/backend/apiClient/upsertApiRecord.ts
@@ -25,7 +25,7 @@ export function sanitizeRecord(record: Partial<RQAPI.ApiClientRecord>) {
       if (sanitizedRecord.data.variables) {
         sanitizedRecord.data.variables = Object.fromEntries(
           Object.entries(sanitizedRecord.data.variables)
-            .filter(([entryKey, variable]) => !!variable && !!(entryKey || "").trim())
+            .filter(([entryKey, variable]) => !!variable && !!(entryKey).trim())
             .map(([entryKey, variable]) => {
               const { localValue, key, ...rest } = variable;
               return [entryKey.trim(), rest];


app/src/backend/apiClient/upsertApiRecord.ts
```diff
@@ -21,14 +21,33 @@ export function sanitizeRecord(record: Partial<RQAPI.ApiClientRecord>) {
...
-            const { localValue, ...rest } = variable;
-            return [key, rest];
-          })
+          Object.entries(sanitizedRecord.data.variables)
+            .map(([entryKey, variable]) => {
+              if (!variable) {
+                return null;
+              }
+              const entryKeyTrimmed = (entryKey || "").trim();
+              const valueKeyTrimmed = (variable?.key || "").toString().trim();
```
nafees87n: Whats the difference between entryKey and valueKey? How come we are able to use one as fallback

app/src/backend/apiClient/upsertApiRecord.ts
```diff
@@ -21,14 +21,33 @@ export function sanitizeRecord(record: Partial<RQAPI.ApiClientRecord>) {
...
+              return [finalKey, rest];
+            })
+            .filter((pair): pair is [string, any] => Array.isArray(pair))
         );
+
+        if (Array.isArray(sanitizedRecord.data.variablesOrder)) {
+          sanitizedRecord.data.variablesOrder = sanitizedRecord.data.variablesOrder
+            .map((k) => (k || "").trim())
+            .filter((k) => k && sanitizedRecord.data?.variables[k]);
+        }
```
nafees87n, line 40: What is the purpose of this? Please add a comment here explaining that

app/src/backend/apiClient/upsertApiRecord.ts
rohitneharabrowserstack: entryKey is the key of each entry in the variables object (the map key).
valueKey comes from the object payload itself variable.key, which can appear when data was imported or previously stored with that embedded key.

We take [finalKey = entryKeyTrimmed || valueKeyTrimmed] so that if one side is empty/whitespace (or missing due to legacy/buggy input), the other can still provide a usable key. This prevents Firestore from seeing empty keys and lets either source define the variable name.

eg:- 
"Veriables":
{
"jjjjj": {
"syncValue": "",
"id": "663aa048-bc25-420b-90d2-c13c6056cc85",
"type": "string",
"key": "",
"isPersisted": true
}
}

app/src/backend/apiClient/upsertApiRecord.ts
rohitneharabrowserstack: Done

app/src/backend/apiClient/upsertApiRecord.ts
nafees87n: > We take [finalKey = entryKeyTrimmed || valueKeyTrimmed] so that if one side is empty/whitespace

I think both the keys should be in sync always. In what cases would they diverge? Correct me if I am wrong

rohitneharabrowserstack: So when u update the key with empty them they diverge

app/src/backend/apiClient/upsertApiRecord.ts
```diff
@@ -21,14 +21,33 @@ export function sanitizeRecord(record: Partial<RQAPI.ApiClientRecord>) {
...
+                return null;
+              }
+              const entryKeyTrimmed = (entryKey || "").trim();
+              const finalKey = entryKeyTrimmed;
+
+              if (!finalKey) {
+                return null; // will be filtered out
+              }
+
+              const { localValue, key: _ignored, ...rest } = variable;
```
nafees87n: Do we still get this key in the object?

app/src/backend/apiClient/upsertApiRecord.ts
```diff
@@ -21,14 +21,33 @@ export function sanitizeRecord(record: Partial<RQAPI.ApiClientRecord>) {
...
+              return [finalKey, rest];
+            })
+            .filter((pair): pair is [string, any] => Array.isArray(pair))
```
nafees87n: I feel its better to first filter out the invalid values and the map to create the desired format. WDYT?

app/src/backend/apiClient/upsertApiRecord.ts
rohitneharabrowserstack: Yes we do get it

app/src/backend/apiClient/upsertApiRecord.ts
rohitneharabrowserstack: Yes makes sense pushed the change

app/src/backend/apiClient/upsertApiRecord.ts
```diff
@@ -21,14 +21,23 @@ export function sanitizeRecord(record: Partial<RQAPI.ApiClientRecord>) {
...
-      // Remove localValue from collection variables
+      // Remove localValue and drop entries with empty keys (from entry or value.key)
       if (sanitizedRecord.data.variables) {
         sanitizedRecord.data.variables = Object.fromEntries(
-          Object.entries(sanitizedRecord.data.variables).map(([key, variable]) => {
-            const { localValue, ...rest } = variable;
-            return [key, rest];
-          })
+          Object.entries(sanitizedRecord.data.variables)
+            .filter(([entryKey, variable]) => !!variable && !!(entryKey || "").trim())
```
nafees87n:
```suggestion
            .filter(([entryKey, variable]) => !!variable && !!(entryKey).trim())
```

This default OR value wont be needed

nafees87n (approved):

PR #4528, opened 2026-03-16
Fixed bug with openApi import issue on local workspace

rohitneharabrowserstack:
Closes issue: 

## 🎥 Demo Video:

**Video/Demo:** 

## ✅ Checklist:

- [ ] Verified UI in browser.

## 🔗 Other references:

##Bug Fixes

Bug 1 — Only 1 of 3 environments imported
Cause: Electron IPC uses ipcMain.once with a channel name derived from the method — e.g. local_sync:/path-createEnvironment. All 3 parallel createEnvironment calls share the same channel name. The first reply consumes the listener; the other 2 have no listener, time out, and fail silently.

Fix (CommonApiClientImporter.tsx): Run environment creation sequentially (not in parallel) when the workspace is a local filesystem instance.

Bug 2 — Multipart form-data body broken
Cause: In getMultipartFormRequestBody inside @requestly/alternative-importers, two mistakes existed side-by-side:

// BEFORE — value computed but thrown away, type never set
propSchema.example ?? propSchema.default ?? '';   // ← result discarded!
formData.push({ id, key, value: "", isEnabled: true });  // always empty, no type

Because type was never set (undefined), the UI condition [record.type === "file"] was always false → file picker never shown. Because [value] was always "" → text field values from the spec's [example][default] were lost.

Fix ([openapi-3.ts]:
// AFTER
const isFileField = propSchema.format === 'binary' || propSchema.format === 'base64';
const value = isFileField ? [] : String(propSchema.example ?? propSchema.default ?? '');
formData.push({ id, key, value, isEnabled: true, type: isFileField ? 'file' : 'text' });

Fix (MultipartFormRowTable.tsx): The file/text <Select> was using defaultValue (uncontrolled — only read on first mount). Changed to [value] (controlled) so it always reflects the actual record state.

<h3 style="line-height: normal; font-size: 13px; font-weight: unset; margin: 0px 0px 8px; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; color: rgb(204, 204, 204); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(24, 24, 24); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Changed Files (final state)</h3>
Repo | File | Fix
-- | -- | --
requestly | CommonApiClientImporter.tsx | Sequential env creation for local FS
requestly | MultipartFormRowTable.tsx | Controlled <Select value=...>
alternative-importers | openapi-3.ts | Correct type + value in multipart body parser

## Summary by CodeRabbit

* **Bug Fixes**
  * Environment imports are more stable for local file system workspaces by handling each environment import sequentially and reporting per-environment outcomes.
  * Fixed multipart form editor behavior so key/value fields remount correctly and reliably reflect edits.

### DIFF
--- a/app/src/features/apiClient/screens/apiClient/components/views/components/request/components/KeyValueTable/MultipartFormRowTable.tsx
+++ b/app/src/features/apiClient/screens/apiClient/components/views/components/request/components/KeyValueTable/MultipartFormRowTable.tsx
@@ -170,7 +170,7 @@ export const MultiEditableCell: React.FC<React.PropsWithChildren<EditableCellPro
 
             {(dataIndex === "key" || (dataIndex === "value" && record.type === FormDropDownOptions.TEXT)) && (
               <SingleLineEditor
-                key={dataIndex === "value" ? `value-${record.type ?? "text"}` : undefined}
+                key={`${dataIndex}-${record.id}-${record.type ?? FormDropDownOptions.TEXT}`}
                 className={`key-value-table-input ${
                   record.isEnabled === false ? "key-value-table-input-disabled" : ""
                 }`}


app/src/features/apiClient/screens/apiClient/components/views/components/request/components/KeyValueTable/MultipartFormRowTable.tsx
```diff
@@ -158,7 +158,7 @@ export const MultiEditableCell: React.FC<React.PropsWithChildren<EditableCellPro
                     label: "File",
                   },
                 ]}
-                defaultValue={record?.type ?? FormDropDownOptions.TEXT}
+                value={record?.type ?? FormDropDownOptions.TEXT}
```
nafees87n: Why this change?

app/src/features/apiClient/screens/apiClient/components/views/components/request/components/KeyValueTable/MultipartFormRowTable.tsx
```diff
@@ -170,6 +170,7 @@ export const MultiEditableCell: React.FC<React.PropsWithChildren<EditableCellPro
 
             {(dataIndex === "key" || (dataIndex === "value" && record.type === FormDropDownOptions.TEXT)) && (
               <SingleLineEditor
+                key={dataIndex === "value" ? `value-${record.type ?? "text"}` : undefined}
```
nafees87n: Don't we have any deterministic prop to be made key?
If we are defaulting to undefined then whats the use of having a key

app/src/features/apiClient/screens/apiClient/components/CommonApiClientImporter/CommonApiClientImporter.tsx
```diff
@@ -59,8 +60,31 @@ export const CommonApiClientImporter: React.FC<CommonApiClientImporterProps> = (
...
+              }) as any
+            ).unwrap();
+            results.push({ status: "fulfilled", value: true });
+          } catch (error) {
+            results.push({ status: "rejected", reason: error });
+          }
+        }
+        return results;
+      }
+
```
nafees87n, line 87: You said create env sequencially but how were they being created bfore this change?

rohitneharabrowserstack: So for desktop app
Before the fix, environments were created in parallel using Promise.allSettled
All 3 createEnvironment calls fired at the same time. Each one goes through Electron IPC to the main process. The IPC bridge registers a one-time listener (ipcMain.once) using a channel name derived from the method something like "local_sync:/path-createEnvironment".

Since all 3 calls used the same channel name, only the first reply consumed the listener. The other 2 had no listener waiting, so they timed out silently resulting in only 1 environment being created.

app/src/features/apiClient/screens/apiClient/components/views/components/request/components/KeyValueTable/MultipartFormRowTable.tsx
rohitneharabrowserstack: It was stray change reverted it

app/src/features/apiClient/screens/apiClient/components/views/components/request/components/KeyValueTable/MultipartFormRowTable.tsx
rohitneharabrowserstack: Done

nafees87n (approved):

Issue #3826, opened 2025-11-11
bug: Rule shows “1 applied filter” even when no filter is applied


dinex-dev:
### Describe your issue?

When viewing the rule in the shared list below, the rule incorrectly displays “1 applied filter” even though no filter has been set.
[-> Shared List](https://app.requestly.io/rules#sharedList/22cc53ef958240a6928233c3913e2565-Mock-Server-Response-GraphQL)

### Repro steps

1. Open the following [shared rule](https://app.requestly.io/rules#sharedList/22cc53ef958240a6928233c3913e2565-Mock-Server-Response-GraphQL)
2. Import it in your workspace
3. Edit the rule
4. Notice that it shows “1 applied filter” even though no filters are configured.

### What Requestly tool were you using?

- [x] Extension
- [x] Desktop-App
- [ ] Android SDK
- [ ] Selenium

### Your Environment

any

### Requestly Version

v25.9.24

#### Shows 1 filter applied
<img width="1470" height="830" alt="Image" src="https://github.com/user-attachments/assets/9296614d-256d-4e86-9610-2ad80e70eb40" />

#### There is no filter
<img width="1470" height="830" alt="Image" src="https://github.com/user-attachments/assets/8897a079-653f-47be-83cc-8bb1083b2db7" />

dhruv-suthar:
@dinex-dev I can work on this bug. Can you assign me?

dinex-dev:
Sure @dhruv-suthar , Assigning this to you

satyam-svg:
I can work on this issue , Can u please assign me ?

priyanshupandey12:
I see this issue was assigned a few days ago. If there’s no activity by the 7-day window and it becomes unassigned, I’d be glad to take it up.

dhruv-suthar:
@priyanshupandey12 I am working on it

priyanshupandey12:
@dhruv-suthar Got it, thanks for the update!

dhruv-suthar:
@dinex-dev @Kanishkrawatt @rohanmathur91 
what are your thoughts on this solution?

https://github.com/user-attachments/assets/3867b7ba-1671-41c5-80db-d70768f1c9ff

dinex-dev:
Hey @dhruv-suthar, this looks good 👍. Please go ahead and raise a PR for your fix.

dhruv-suthar:
@dinex-dev done raised PR. can you review it?

Dinesht04:
[...]

PR #4688, opened 2026-05-13
Fix empty source filter badge count

darshan-Jahagirdar:
## Summary
- fixes the source filter badge count so empty filter values do not show as applied filters
- keeps the existing page URL exclusion from the badge count
- handles empty arrays, blank strings, empty nested objects, and request payload filters without key/value data

Fixes #3826

## Validation
- `git diff --check`
- Node smoke test for filter-count cases covering empty objects, empty arrays, request payload, page URL, and applied filters

Note: local app ESLint was not available in this checkout; no paid services were used.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate active source filter badge: counts only truly applied filters, ignores PAGE_URL, and correctly handles strings, arrays, nested objects, and request data entries to prevent inflated counts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/requestly/requestly/pull/4688)

### DIFF
--- a/app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/RequestSourceRow/index.js
+++ b/app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/RequestSourceRow/index.js
@@ -26,6 +26,34 @@ import "./RequestSourceRow.css";
 
 const { Text } = Typography;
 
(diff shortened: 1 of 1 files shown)


darshan-Jahagirdar:
Addressed the CodeRabbit finding: the REQUEST_DATA branch now trims both key and value before computing the filter badge count, so whitespace-only payload fields are treated as empty.

I kept the change limited to the REQUEST_DATA branch; the other filter-count branches are unchanged.

PR #4723, opened 2026-05-21
fix: ignore empty source filters in badge count

adriangmweb:
## Summary
- Count only source filters that have an actual configured value
- Ignore empty objects, empty arrays, arrays with empty entries, blank strings, and nullish values when rendering the advanced filters badge
- Require request payload filters to have both key and value before counting them as applied
- Keep page URL excluded from the advanced filters count
- Move the filter-counting logic into a small utility with regression coverage

Fixes #3826

## Testing
- Added unit coverage for empty/default filters, arrays with empty entries, meaningful filters, incomplete requestPayload defaults, pageUrl exclusion, and legacy object-shaped filters
- Verified the helper logic locally with a standalone Node check because this fresh clone does not have node_modules installed

## Demo
- No video attached because this is a small badge-counting logic fix; the regression cases are covered by the added unit tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate counting and activation detection of request-source filters, improving the displayed filter badges.
  * Correct handling of both legacy and upgraded filter formats so counts remain reliable.
  * Excludes page URL from advanced-filter totals to prevent inflated counts.

* **Tests**
  * Added tests validating filter detection and advanced-filter counting behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/requestly/requestly/pull/4723?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/RequestSourceRow/index.js
+++ b/app/src/components/features/rules/RulePairs/Pairs/Rows/RowsMarkup/RequestSourceRow/index.js
@@ -92,14 +92,13 @@ const RequestSourceRow = ({ rowIndex, pair, pairIndex, ruleDetails, isInputDisab
 
   const getFilterCount = useCallback(
     (pairIndex) => {
-      const copyOfCurrentlySelectedRule = JSON.parse(JSON.stringify(currentlySelectedRuleData));
       const sourceFilters = currentlySelectedRuleData.pairs[pairIndex].source.filters;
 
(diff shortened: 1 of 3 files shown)


adriangmweb:
Hi maintainers, this is ready for review.

I kept the change scoped to the source filter badge count and added regression coverage for:
- empty/default filter objects
- arrays with only empty entries, e.g. [""], [{}], [null]
- incomplete requestPayload filters where only key or value is set
- pageUrl exclusion and legacy object-shaped source filters

All checks and CLA are passing. Happy to adjust if you prefer a different utility shape.

